PR #11, opened 2021-05-12
feat: LBP pallet

### LINKED ISSUE #25: LBP
PR #11

Roznovjak:
## Description
Implement liquidity bootstrapping pool pallet

closes #25


pallets/lbp/Cargo.toml
```diff
@@ -0,0 +1,68 @@
+[package]
+authors = ['GalacticCouncil']
+description = 'HydraDX Liquidity Bootstrapping Pool Pallet'
+edition = '2018'
+homepage = 'https://github.com/galacticcouncil/hydradx-node'
```
enthusiastmartin, line 5: basilisk ?

pallets/lbp/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+license = 'Unlicense'
+name = 'pallet-lbp'
+repository = 'https://github.com/galacticcouncil/hydradx-node'
```
enthusiastmartin, line 8: basilisk ?

pallets/lbp/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+version = '3.0.0'
```
enthusiastmartin, line 9: Why not to start with version 1 ?

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,964 @@
...
+
+		Overflow,
+	}
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Create new LBP pool
+		/// who, asset a, asset b, amount_a, amount_b
+		CreatePool(T::AccountId, AssetId, AssetId, BalanceOf<T>, BalanceOf<T>),
```
enthusiastmartin, line 237: As discussed - events should be names with past tense. There is a mix of present and past tenses.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,964 @@
...
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Create pool errors
+		CannotCreatePoolWithSameAssets,
```
enthusiastmartin, line 182: Add error descriptions for each error. It is used on the FE side.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,964 @@
...
+		/// Sell token
+		/// who, asset in, asset out, amount, sale price
+		SellExecuted(T::AccountId, AssetId, AssetId, BalanceOf<T>, BalanceOf<T>),
+
+		/// Buy token
+		/// who, asset out, asset in, amount, buy price
+		BuyExecuted(T::AccountId, AssetId, AssetId, BalanceOf<T>, BalanceOf<T>),
+
+		Paused(T::AccountId),
+		Unpaused(T::AccountId),
```
enthusiastmartin, line 264: In these events - PoolId would be useful too ?

enthusiastmartin (commented):
First very quick review. Only some obvious ones. 

There are still some TODOs in code  - so i will wait for final version before continuing.

enthusiastmartin (commented):
Few more comments. I still have to finish it off.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1017 @@
...
+			}
+
+			if let Some((w1, w2)) = final_weights {
+				pool_data.final_weights = (w1, w2);
+			}
+
+			Self::validate_pool_data(&pool_data)?;
+
+			<PoolData<T>>::insert(&pool_id, &pool_data);
+			Self::deposit_event(Event::PoolUpdated(who, pool_id));
```
enthusiastmartin, line 447:
```suggestion
			<PoolData<T>>::try_mutate(pool_id.clone(), |pool| -> DispatchResult {
				let now = <frame_system::Pallet<T>>::block_number();

				ensure!(
					pool.start == Zero::zero() || now < pool.start,
					Error::<T>::SaleStarted
				);

				pool.start = start.unwrap_or(pool.start);
				pool.end = end.unwrap_or(pool.end);
				pool.initial_weights = initial_weights.unwrap_or(pool.initial_weights);
				pool.final_weights = final_weights.unwrap_or(pool.final_weights);

				Self::validate_pool_data(&pool)?;
				Self::deposit_event(Event::PoolUpdated(who, pool_id));

				Ok(())

			})?;

```

Something like this.

But the owner check is missing. Hover it is not tested! because all tests are passing with this change.!

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1017 @@
...
+			origin: OriginFor<T>,
+			pool_id: PoolId<T>,
+			start: Option<T::BlockNumber>,
+			end: Option<T::BlockNumber>,
+			initial_weights: Option<((AssetId, LBPWeight), (AssetId, LBPWeight))>,
+			final_weights: Option<((AssetId, LBPWeight), (AssetId, LBPWeight))>,
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+
+			ensure!(<PoolOwner<T>>::contains_key(&pool_id), Error::<T>::PoolNotFound);
```
enthusiastmartin, line 416: As mentioned in the other comment - this owner check is not tested

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1017 @@
...
+			let (mut reserve_a, mut reserve_b) = Self::pool_balances(&pool_id);
+			let (asset_a, asset_b) = Self::pool_assets(&pool_id);
+
+			let pool_data = Self::pool_data(&pool_id);
+
+			let now = <frame_system::Pallet<T>>::block_number();
+			ensure!(
+				pool_data.start == Zero::zero() || now < pool_data.start,
+				Error::<T>::SaleStarted
+			);
```
enthusiastmartin, line 526: You can move this before reading balances from the storage. So in case of error - storage calls to get balances and assets are not executed .

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1017 @@
...
+
+			if !amount_a.is_zero() {
+				ensure!(
+					T::MultiCurrency::free_balance(asset_a, &who) >= amount_a,
+					Error::<T>::InsufficientAssetBalance
+				);
+
+				reserve_a = reserve_a.checked_add(amount_a).ok_or(Error::<T>::LiquidityOverflow)?;
```
enthusiastmartin, line 534: I believe that this can somehow reorganized so the call to get pool balances from storage are executed only if the free balance check is ok.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1017 @@
...
+
+			ensure!(!Self::is_pool_running(&pool_data), Error::<T>::SaleNotEnded);
+
+			if !amount_a.is_zero() {
+				reserve_a = reserve_a.checked_sub(amount_a).ok_or(Error::<T>::LiquidityUnderflow)?;
+			}
+
+			if !amount_b.is_zero() {
+				reserve_b = reserve_b.checked_sub(amount_b).ok_or(Error::<T>::LiquidityUnderflow)?;
+			}
```
enthusiastmartin, line 587: Same applies here what was mentioned for the add_liquduity. It can reorganized so check are done prior to making any storage calls.

enthusiastmartin (commented):
Few more comments. Sorry that it comes in chunks but there is lot of check and think about.

I will carry on again later.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1017 @@
...
+
+		#[pallet::weight(<T as Config>::WeightInfo::pause_pool())]
+		pub fn pause_pool(origin: OriginFor<T>, pool_id: PoolId<T>) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+
+			ensure!(<PoolOwner<T>>::contains_key(&pool_id), Error::<T>::PoolNotFound);
+
+			Self::ensure_pool_ownership(&who, &pool_id)?;
+
+			let mut pool_data = Self::pool_data(&pool_id);
```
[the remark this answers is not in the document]
Roznovjak: Is there a reason why to perform the owner check inside _try_mutate_? I'm just curious.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,998 @@
...
+				Ok(weights) => {
+					if weights.0 .0 == asset_a {
+						weight_a = weights.0 .1;
+						weight_b = weights.1 .1;
+					} else {
+						weight_a = weights.1 .1;
+						weight_b = weights.0 .1;
+					}
+
+					<PoolData<T>>::insert(&pool_id, &pool_data);
```
enthusiastmartin, line 939: I wonder why this function changes the pool? 

it should only calculate spot price.

I dont think that this should make storage changes - if yes, it probably needs some refactoring.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,998 @@
...
+			false => None,
+		}
+	}
+
+	/// Calculate spot price for given assets and amount and update pool weights if necessary
+	///
+	/// Provided assets must exist in the pool. Panic if an asset does not exist in the pool.
+	///
+	/// Return 0 if calculation overflows or weights calculation overflows.
+	fn get_spot_price_unchecked(asset_a: AssetId, asset_b: AssetId, amount: BalanceOf<T>) -> BalanceOf<T> {
```
enthusiastmartin, line 908: No tests for this. 

But as mentioned in other comment - it probably needs some refactor beause it does more that the name suggests.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,998 @@
...
+	}
+
+	fn get_pool_assets(pool_account_id: &T::AccountId) -> Option<Vec<AssetId>> {
+		match <PoolAssets<T>>::contains_key(pool_account_id) {
+			true => {
+				let assets = Self::pool_assets(pool_account_id);
+				Some(vec![assets.0, assets.1])
+			}
+			false => None,
+		}
```
enthusiastmartin, line 900:
```suggestion
		if <PoolAssets<T>>::contains_key(pool_account_id) {
			let assets = Self::pool_assets(pool_account_id);
			Some(vec![assets.0, assets.1])
		}else{
			None
		}
```

Also add test case to test with non existing pool account id

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,998 @@
...
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Pool assets can not be the same
+		CannotCreatePoolWithSameAssets,
+		/// Initial liquidity should be non-zero
+		CannotCreatePoolWithZeroLiquidity,
```
enthusiastmartin, line 202: No test for this error.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,998 @@
...
+		/// Account is not a pool owner
+		NotOwner,
+		/// Sale already started
+		SaleStarted,
+		/// Sale is still in progress
+		SaleNotEnded,
+		/// Sale is not running
+		SaleIsNotRunning,
+		/// Provided data are not valid
+		InvalidData,
```
enthusiastmartin, line 212: This error is not used. So it can be removed.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,998 @@
...
+		/// Pool is not paused
+		PoolIsNotPaused,
+		/// Sale duration is too long
+		MaxSaleDurationExceeded,
+		/// Liquidity being added should not be zero
+		CannotAddZeroLiquidity,
+		/// Can not remove zero liquidity
+		CannotRemoveZeroLiquidity,
+		/// Overflow after adding liquidity
+		LiquidityOverflow,
```
enthusiastmartin, line 230: No tests.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,998 @@
...
+		/// Asset balance too low
+		InsufficientAssetBalance,
+		/// Pool does not exist
+		PoolNotFound,
+		/// Pool has been already created
+		PoolAlreadyExists,
+		/// Invalid block number
+		InvalidBlockNumber,
+		/// Calculation error
+		WeightCalculationError,
```
enthusiastmartin, line 242: No tests

pallets/lbp/src/tests.rs
```diff
@@ -0,0 +1,1680 @@
...
+		assert_eq!(LBPPallet::pool_deposit(&ACA_DOT_POOL_ID), POOL_DEPOSIT);
+
+		assert_eq!(LBPPallet::get_pool_assets(&ACA_DOT_POOL_ID).unwrap(), vec![ACA, DOT]);
+
+		// verify that `last_weight_update`, `last_weights` and `paused` fields are correctly initialized
+		let updated_pool_data = LBPPallet::pool_data(ACA_DOT_POOL_ID);
+		assert_eq!(updated_pool_data.last_weight_update, 0);
+		assert_eq!(updated_pool_data.last_weights, ((ACA, 20), (DOT, 80)));
+		assert_eq!(updated_pool_data.paused, true);
+
```
enthusiastmartin, line 254: it might be good check all the storage changes after pool created - so PoolOwner, Balacnes are missing here.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,998 @@
...
+
+		#[pallet::weight(<T as Config>::WeightInfo::update_pool_data())]
+		#[transactional]
+		pub fn update_pool_data(
+			origin: OriginFor<T>,
+			pool_id: PoolId<T>,
+			start: Option<T::BlockNumber>,
+			end: Option<T::BlockNumber>,
+			initial_weights: Option<((AssetId, LBPWeight), (AssetId, LBPWeight))>,
+			final_weights: Option<((AssetId, LBPWeight), (AssetId, LBPWeight))>,
```
enthusiastmartin, line 392: Perhaps add a check to verify if at least one thing is not None, so there is actually something to update.

pallets/lbp/src/tests.rs
```diff
@@ -0,0 +1,1680 @@
...
+			None,
+			Some(((ACA, 10), (DOT, 90))),
+			None,
+		));
+
+		// verify changes
+		let updated_pool_data = LBPPallet::pool_data(ACA_DOT_POOL_ID);
+		assert_eq!(updated_pool_data.start, 15);
+		assert_eq!(updated_pool_data.end, 20);
+
```
enthusiastmartin, line 415: the update of initial weights and final weights is not tested. 

i would also add a simple test to verify that if you update one ting (eg. start) - the rest pool parameters are preserved and not changed.

pallets/lbp/src/tests.rs
```diff
@@ -0,0 +1,1680 @@
...
+			user_balance_b_after,
+			user_balance_b_before.saturating_add(balance_b_before)
+		);
+
+		let user_balance_hdx_after = Currency::reserved_balance(HDX, &ALICE);
+		assert_eq!(
+			user_balance_hdx_after,
+			user_balance_hdx_before.saturating_sub(POOL_DEPOSIT)
+		);
+
```
enthusiastmartin, line 1050: I would also add check that storage was cleared ( assets, pool data, owner, balances, deposit).

enthusiastmartin (commented):
Few comments on the  tests.

pallets/lbp/src/lib.rs
Roznovjak: What to do when all fields are set to None? Throw the PoolUpdated event and skip the rest or throw an error?

pallets/lbp/src/lib.rs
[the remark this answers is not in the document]
enthusiastmartin: > 
> 
> Is there a reason why to perform the owner check inside _try_mutate_? I'm just curious.

I updated my suggestions to use try_mutate_exists instead.

So now, you can use it and do the check owner inside. In this way, there is no need to call contains key to check if pool exists.

pallets/lbp/src/lib.rs
enthusiastmartin: > 
> 
> What to do when all fields are set to None? Throw the PoolUpdated event and skip the rest or throw an error?

I would just return an error - something saying that there is nothing to update.

pallets/lbp/src/lib.rs
Roznovjak: Calculated weights are stored so there is no need to calculate them again in the same block. I have moved the _insert_ method to _update_weights_ function.

lumir-mrkva (changes requested):
I don't know why token `reserve` is necessary. Spot price with current "future" weight should be calculated without actually changing the weight in storage.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1017 @@
...
+
+		/// Mapping of asset pairs to unique pool identities
+		type AssetPairPoolId: AssetPairPoolIdFor<AssetId, PoolId<Self>>;
+
+		#[pallet::constant]
+		/// Storage fee
+		type PoolDeposit: Get<BalanceOf<Self>>;
+
+		/// Trading fee rate
+		type ExchangeFee: Get<fee::Fee>;
```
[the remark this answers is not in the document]
Roznovjak: removed from the config. Not relevant anymore

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+				T::MultiCurrency::free_balance(asset_b.id, &pool_owner) >= asset_b.amount,
+				Error::<T>::InsufficientAssetBalance
+			);
+
+			let pool_data = Pool::new(asset_a, asset_b, sale_duration, weight_curve, pausable);
+			Self::validate_pool_data(&pool_data)?;
+
+			let deposit = T::PoolDeposit::get();
+
+			T::MultiCurrency::reserve(T::NativeAssetId::get(), &pool_owner, deposit)?;
```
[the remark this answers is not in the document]
Roznovjak: do we want to set the deposit value to zero, or to remove it completely?

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+		let weight_a: LBPWeight;
+		let weight_b: LBPWeight;
+		// update weight or reuse last if update is not necessary
+		if pool_data.last_weight_update == <frame_system::Pallet<T>>::block_number() {
+			let w = Self::get_weights_in_order(&pool_data, asset_a);
+
+			weight_a = w.0;
+			weight_b = w.1;
+		} else {
+			match Self::update_weights(&pool_id, &mut pool_data) {
```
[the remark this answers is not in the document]
Roznovjak: fixed

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+		Paused(T::AccountId, PoolId<T>),
+
+		/// Pool was unpaused. [who, pool_id]
+		Unpaused(T::AccountId, PoolId<T>),
+	}
+
+	/// Paid deposit. Returned when a pool is destroyed and removed from the storage.
+	#[pallet::storage]
+	#[pallet::getter(fn pool_deposit)]
+	pub type PoolDeposit<T: Config> = StorageMap<_, Blake2_128Concat, PoolId<T>, BalanceOf<T>, ValueQuery>;
```
enthusiastmartin, line 284: If nothing else is going to be done with this PoolDeposit - it is actually useless. Because currently it only stores same amount for every pool ( pallet's config option), and it is only removed from storage.

So you can simply remove it completely. Because you wont' allow to create a pool if `T::MultiCurrency::reserve` fails.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+			amount_a: (AssetId, BalanceOf<T>),
+			amount_b: (AssetId, BalanceOf<T>),
+		) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+
+			let (asset_a, asset_b) = (amount_a.0, amount_b.0);
+			let (amount_a, amount_b) = (amount_a.1, amount_b.1);
+
+			let asset_pair_id = T::AssetPairPoolId::from_assets(asset_a, asset_b);
+			ensure!(pool_id == asset_pair_id, Error::<T>::InvalidAsset);
```
enthusiastmartin, line 466: I dont see a point in constructing this pair id and performing the check.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
```
enthusiastmartin, line 525: Same here as for add liquidity. Why this has to be done ?

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+			<PoolData<T>>::try_mutate_exists(pool_id.clone(), |maybe_pool| -> DispatchResultWithPostInfo {
+				// check existence of the pool
+				let mut pool = maybe_pool.as_mut().ok_or(Error::<T>::PoolNotFound)?;
+
+				Self::ensure_pool_ownership(&who, &pool_id)?;
+
+				ensure!(pool.paused, Error::<T>::PoolIsNotPaused);
+
+				let now = <frame_system::Pallet<T>>::block_number();
+				ensure!(pool.end > now, Error::<T>::CannotUnpauseEndedPool);
```
enthusiastmartin, line 443: Should not this be `>=` ?

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+				ensure!(
+					start.is_some() || end.is_some() || initial_weights.is_some() || final_weights.is_some(),
+					Error::<T>::NothingToUpdate
+				);
+
+				Self::ensure_pool_ownership(&who, &pool_id)?;
+
+				let now = <frame_system::Pallet<T>>::block_number();
+
+				ensure!(pool.start == Zero::zero() || now < pool.start, Error::<T>::SaleStarted);
```
enthusiastmartin, line 389: I would consider creating little helper function to check the start and end since it used multiple times and it is not a simple  check. 

There are is_pool_running, is_sale_running - so something along those.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+
+				pool.start = start.unwrap_or(pool.start);
+				pool.end = end.unwrap_or(pool.end);
+				if let Some(initial_weights) = initial_weights {
+					pool.initial_weights = Self::get_weights_in_order(pool, initial_weights)?;
+				}
```
enthusiastmartin, line 395:
```suggestion
				pool.initial_weights = initial_weights.map_or(Ok(pool.initial_weights),  |w| Self::get_weights_in_order(pool, w))?;
```

If you want to go with the theme `unwrap_or` ;-)

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+				if let Some(final_weights) = final_weights {
+					pool.final_weights = Self::get_weights_in_order(pool, final_weights)?;
```
enthusiastmartin, line 397: same here.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+				// check existence of the pool
+				let mut pool = maybe_pool.as_mut().ok_or(Error::<T>::PoolNotFound)?;
+
+				Self::ensure_pool_ownership(&who, &pool_id)?;
+
+				ensure!(pool.pausable, Error::<T>::PoolIsNotPausable);
+				ensure!(!pool.paused, Error::<T>::CannotPausePausedPool);
+
+				let now = <frame_system::Pallet<T>>::block_number();
+				ensure!(pool.end > now, Error::<T>::CannotPauseEndedPool);
```
enthusiastmartin, line 421: as mentioned - perhaps a little helper for this would be useful. 

And if feel it should be `>='.

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+			let (asset_a, asset_b) = pool_data.assets;
+
+			let amount_a = T::MultiCurrency::free_balance(asset_a, &pool_id);
+			let amount_b = T::MultiCurrency::free_balance(asset_b, &pool_id);
+
+			T::MultiCurrency::transfer(asset_a, &pool_id, &who, amount_a)?;
+			T::MultiCurrency::transfer(asset_b, &pool_id, &who, amount_b)?;
+
+			let deposit = Self::pool_deposit(&pool_id);
+			T::MultiCurrency::unreserve(T::NativeAssetId::get(), &who, deposit);
```
enthusiastmartin, line 588: I would also check if all deposit has been unreserved. 

You can do so by checking the return value.  And fire an event if there is discrepancy.

pallets/lbp/src/lib.rs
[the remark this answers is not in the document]
lumir-mrkva: i would personally remove it completely, do you think there will be a need for it in the future?

pallets/lbp/src/benchmarking.rs
```diff
@@ -0,0 +1,323 @@
...
+		LBP::<T>::create_pool(RawOrigin::Root.into(), caller.clone(), asset_a, asset_b, duration, WeightCurveType::Linear, true)?;
+		ensure!(PoolData::<T>::contains_key(&pool_id), "Pool does not exist.");
+
+		LBP::<T>::unpause_pool(RawOrigin::Signed(caller.clone()).into(), pool_id.clone())?;
+		let pool_data = LBP::<T>::pool_data(&pool_id);
+		assert_eq!(pool_data.paused, false);
+
+		System::<T>::set_block_number(12u32.into());
+
+	}: _(RawOrigin::Signed(caller.clone()), asset_in, asset_out, amount, max_limit)
```
enthusiastmartin, line 258: does this benchmark take into account the weight update when validating ?  Because that would the worst case here.

pallets/lbp/src/lib.rs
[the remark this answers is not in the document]
Roznovjak: removed

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,987 @@
...
+			<PoolOwner<T>>::insert(&pool_id, &pool_owner);
+			<PoolAssets<T>>::insert(&pool_id, &(asset_a.id, asset_b.id));
+			<PoolData<T>>::insert(&pool_id, &pool_data);
+
+			T::MultiCurrency::transfer(asset_a.id, &pool_owner, &pool_id, asset_a.amount)?;
+			T::MultiCurrency::transfer(asset_b.id, &pool_owner, &pool_id, asset_b.amount)?;
+
+			<PoolBalances<T>>::insert(&pool_id, &(asset_a.amount, asset_b.amount));
+
+			Self::deposit_event(Event::PoolCreated(
```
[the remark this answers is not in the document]
Roznovjak: Not sure what you mean. Do you mean to use the `Pool` type in the `PoolCreated` event?

pallets/lbp/src/lib.rs
Roznovjak: Regarding the storage, (afaik) this pattern is commonly used in substrate pallets. This way it's possible to change the deposit value in the future, keep track of previous deposits, and charge the difference if necessary.
But in the case of LBP, the deposit functionality has been removed.

pallets/lbp/src/lib.rs
Roznovjak: This check is used to ensure that we are not using this method to add wrong asset to the pool specified by pool_id param (e.g. ETH to HDX/ACA pool).

Roznovjak: It is probably not necessary to expect both the `pool_id` parameter and asset pair. `Pool_id` can be removed. What do you think?

pallets/lbp/src/lib.rs
Roznovjak: Didn't know that `unreserve` can "fail". Thanks. But this issue is not relevant anymore. Closing...

pallets/lbp/src/benchmarking.rs
Roznovjak: yes, weights are updated because there is only one trade event in the block (weights are updated prior the first trade, in the same extrinsic), and the block number is different than the `start` block. But I'm going to double-check it, somehow. Just to be sure.

pallets/lbp/src/lib.rs
enthusiastmartin: Well, it was not clear for me the purpose of this check. Now that i know - why not just to check PoolData's assets?

Roznovjak: we verify that the provided pool_id and assets target the same pool.

pallets/lbp/src/lib.rs
Roznovjak: I have decided to remove the LiquidityOverflow error because I was not able to create a test for it (we can't simulate such a scenario)

pallets/lbp/src/lib.rs
Roznovjak: probably yes

runtime/src/lib.rs
```diff
@@ -156,6 +157,7 @@ impl Filter<Call> for BaseFilter {
 			| Call::Sudo(_) => true,
 
 			Call::XYK(_)
+			| Call::LBP(_)
```
lumir-mrkva, line 160: LBP calls can be allowed from start

Roznovjak: done

pallets/lbp/src/lib.rs
Roznovjak: `pool_id` param has been removed

enthusiastmartin (commented):
It looks to me now. I have no more comments. 

I'd suggest for @lumir-mrkva to do one last review before merging.

pallets/lbp/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+default-features = false
+features = ['derive']
+package = 'parity-scale-codec'
+version = '2.0.0'
+
+[dependencies]
+primitive-types = {default-features = false, version = '0.8.0'}
+serde = {features = ['derive'], optional = true, version = '1.0.101'}
+
+hydra-dx-math = {default-features = false, version = "1.0.0", git = "https://github.com/galacticcouncil/hydraDX-math", branch="feat/lbp"}
```
lumir-mrkva, line 28: merge math 1st https://github.com/galacticcouncil/HydraDX-math/pull/17 and correct dependency version

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1059 @@
...
+			);
+		} else {
+			ensure!(
+				amount <= asset_out_reserve / MAX_OUT_RATIO,
+				Error::<T>::MaxOutRatioExceeded
+			);
+		}
+
+		let (amount_in, amount_out, fee_asset, fee_amount) = if trade_type == TradeType::Sell {
+			let token_amount_out = hydra_dx_math::lbp::calculate_out_given_in(
```
lumir-mrkva, line 829: fee should be part of this calculation, so the frontend can return the correct value, you can return transfer fee as second return value for transfer

pallets/lbp/src/lib.rs
```diff
@@ -0,0 +1,1059 @@
...
+					(weights.0, weights.1)
+				} else {
+					// swap weights if assets are in different order
+					(weights.1, weights.0)
+				}
+			}
+			Err(_) => return BalanceOf::<T>::zero(),
+		};
+
+		hydra_dx_math::lbp::calculate_spot_price(asset_a_reserve, asset_b_reserve, weight_in, weight_out, amount)
```
lumir-mrkva, line 1003: spot price ignores fee? this would be solved by earlier comment

lumir-mrkva (changes requested):
- [ ] move fee calculations to math
- [x] release math
- [ ] bump runtime version

pallets/lbp/src/lib.rs
Roznovjak: As discussed, we are not going to add transfer fee to hydra_dx_math package because it brings new dependencies.

pallets/lbp/src/lib.rs
Roznovjak: transfer fee does not make sense for the spot price

pallets/lbp/Cargo.toml
```diff
@@ -0,0 +1,68 @@
...
+orml-traits = {default-features = false, version = "0.4.1-dev"}
+
+## Substrate dependencies
+frame-benchmarking = {default-features = false, optional = true, version = '3.0.0'}
+frame-support = {default-features = false, version = '3.0.0'}
+frame-system = {default-features = false, version = '3.0.0'}
+frame-system-benchmarking = {default-features = false, optional = true, version = '3.0.0'}
+sp-std = {default-features = false, version = '3.0.0'}
+
+#[dev-dependencies]
```
lumir-mrkva, line 44: try to enable dev dependencies

```suggestion
[dev-dependencies]
```

pallets/lbp/src/weights.rs
```diff
@@ -0,0 +1,90 @@
...
+	fn sell() -> Weight;
+	fn buy() -> Weight;
+}
+
+/// Weights for LBP pallet using the hydraDX node and recommended hardware.
+pub struct HydraWeight<T>(PhantomData<T>);
+
+impl<T: frame_system::Config> WeightInfo for HydraWeight<T> {
+	fn create_pool() -> u64 {
+		0
```
lumir-mrkva, line 25: @enthusiastmartin please benchmark

lumir-mrkva (approved):

PR #39, opened 2021-05-26
chore: pin orml to specific revision

enthusiastmartin:
## Description
Basilisk uses orml dependencies from master branch. Since there was a change to use more recent substrate version  from different branch ( polkadot-v0.9.2) - doing any cargo update will result in having 2 different versions of substrate. 

Basilisk is currently built with substrate from branch polkadot-v0.9.1

Therefore, this PR pins the orml to specific revision where the same version of substrate is used.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -772,9 +772,9 @@ dependencies = [
 
 [[package]]
 name = "blake3"
-version = "0.3.7"
+version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e9ff35b701f3914bdb8fad3368d822c766ef2858b2583198e41639b936f09d3f"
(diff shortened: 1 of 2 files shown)


lumir-mrkva:
@martinfridrich why we just dont upgrade to 0.9.2? well have to upgrade to 0.9.3 anyway soon

enthusiastmartin:
> @martinfridrich why we just dont upgrade to 0.9.2? well have to upgrade to 0.9.3 anyway soon

yes, sure. we can update. but as always,i'd expect some effort and this was jus to make sure that basilisk works ok even when it needs some updates. 

i would say that it would be probably good to have orml pinned some specific revision and update when necessary.

lumir-mrkva (approved):

PR #59, opened 2021-06-09
feat: utility

green-jay:
Fixes: #50


lumir-mrkva (changes requested):
increment runtime version pls

lumir-mrkva (approved):

Issue #69, opened 2021-06-18
HydraDX LBP claims


lumir-mrkva:
- [ ] vested claim (?)
- [ ] data set in chainspec

lumir-mrkva:
only claimed HDX is airdropped to the Basilisk

PR #71, opened 2021-06-24
feat: vesting

green-jay:
## Description
Add vested claims for LBP and other crowdfund mechanisms where vesting is needed

## Related Issue
Fixes: #69


node/src/chain_spec.rs
```diff
@@ -103,6 +118,7 @@ pub fn testnet_parachain_config(para_id: ParaId) -> Result<ChainSpec, String> {
 				//technical committee
 				vec![hex!["30035c21ba9eda780130f2029a80c3e962f56588bc04c36be95a225cb536fb55"].into()],
 				hex!["30035c21ba9eda780130f2029a80c3e962f56588bc04c36be95a225cb536fb55"].into(), // SAME AS ROOT
+				get_vesting_config_for_test(),
```
lumir-mrkva, line 121: testnet = westend shedules wil be empty

node/src/chain_spec.rs
```diff
@@ -363,7 +382,7 @@ fn parachain_genesis(
 			members: tech_committee_members,
 			phantom: Default::default(),
 		},
-		vesting: VestingConfig { vesting: vec![] },
+		vesting: VestingConfig { vesting: vesting_list },
```
lumir-mrkva, line 385:
```suggestion
		vesting: VestingConfig { vesting: vec![] },
```

runtime/Cargo.toml
```diff
@@ -49,6 +49,11 @@ orml-tokens = { default-features = false, version = "0.4.1-dev" }
 orml-traits = { default-features = false, version = "0.4.1-dev" }
 #orml-xtokens = { default-features = false, version = "0.4.1-dev" }
 #orml-xcm-support = { default-features = false, version = "0.4.1-dev" }
+## version from 02/2021 - working
+orml-vesting = {default-features = false, version = '0.4.0'}
+## version 05/2021 - breaks build
+#orml-vesting = { git = 'https://github.com/open-web3-stack/open-runtime-module-library', rev = '7c9fd1c7491f8933a589985b7ed434519beeb8bb' }
```
lumir-mrkva, line 55: 0.4.1-dev not working? doesn't this introduce some dependency hell with duplicated packages?

runtime/src/lib.rs
```diff
@@ -642,6 +630,49 @@ impl pallet_session::Config for Runtime {
 	type WeightInfo = ();
 }
 
+parameter_types! {
+	pub IntergalacticAccounts: Vec<AccountId> = vec![
+		hex_literal::hex!["0abad795adcb5dee45d29528005b1f78d55fc170844babde88df84016c6cd14d"].into(),
```
lumir-mrkva, line 635: how you can use vesting in dev environment when this account is only allowed one, sudo?

runtime/src/lib.rs
```diff
@@ -642,6 +630,49 @@ impl pallet_session::Config for Runtime {
...
+	pub const MaxVestingSchedules: u32 = 100;
+}
+
+impl orml_vesting::Config for Runtime {
+	type Event = Event;
+	type Currency = Balances;
+	type MinVestedTransfer = MinVestedTransfer;
+	type VestedTransferOrigin = EnsureIntergalactic;
+	type WeightInfo = ();
+	//type MaxVestingSchedules = MaxVestingSchedules;
```
lumir-mrkva, line 673:
```suggestion
	type MaxVestingSchedules = MaxVestingSchedules;
```

is this enforced or is it only used for benchmarks?

runtime/src/lib.rs
```diff
@@ -642,6 +630,49 @@ impl pallet_session::Config for Runtime {
...
+	}
+
+	#[cfg(feature = "runtime-benchmarks")]
+	fn successful_origin() -> Origin {
+		Origin::from(RawOrigin::Signed(Default::default()))
+	}
+}
+
+parameter_types! {
+	pub MinVestedTransfer: Balance = BSX / 100_000;
```
lumir-mrkva, line 663:
```suggestion
	pub MinVestedTransfer: Balance = BSX * 100_000;
```

lumir-mrkva (changes requested):

runtime/src/lib.rs
lumir-mrkva: hmm, i dont see this filter applied anywhere

lumir-mrkva: not sure if sudo will do, this is not filtering origin of funds but the transaction, so only that specific wallet can call it

runtime/src/lib.rs
green-jay: this type was added in 0.4.1-dev

runtime/src/lib.rs
green-jay: I think this is going to be transferred from treasury (like a regular claim) so RootOrTreasury might be the best candidate for origin

runtime/src/lib.rs
green-jay: this is just the minimum for the vested_transfer call defined by the caller

green-jay:
vested transfer and setting vesting schedules tested successfully

runtime/src/lib.rs
lumir-mrkva: i assume this is safety mechanism to prevent spamming out limited number of vestings u can have at the sam etime

runtime/src/lib.rs
lumir-mrkva: depends, i assume that hydra dx forked balances will be minted, same as the lbp claims

lumir-mrkva: i'll rephrase question: this filter means that you can vest tokens only by call from intergalacic address?

green-jay: yes that is correct. not only that! moreover the balance is transferred from the callers account so the caller has to have enough balances. For now, I changed the origin to RootOrTreasury.

runtime/src/lib.rs
green-jay: yes, but depends, if only a privileged account can call vested_transfer it should be no problem?

lumir-mrkva (approved):

PR #91, opened 2021-07-05
feat: add trading limits to xyk pool

enthusiastmartin:
Added trading limits.

### DIFF
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -151,8 +151,9 @@ pub mod pallet {
 		SellAssetAmountInvalid, // no tests
 		/// Overflow
 		BuyAssetAmountInvalid, // no tests
+
 		/// Overflow
-		FeeAmountInvalid, // no tests
+		FeeAmountInvalid,
 
 		/// Overflow
 		CannotApplyDiscount,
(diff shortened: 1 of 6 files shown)


pallets/xyk/src/lib.rs
```diff
@@ -150,6 +153,7 @@ pub mod pallet {
 		BuyAssetAmountInvalid, // no tests
 		/// Overflow
 		FeeAmountInvalid, // no tests
```
jak-pan, line 155: still true?

pallets/xyk/src/lib.rs
```diff
@@ -150,6 +153,7 @@ pub mod pallet {
 		BuyAssetAmountInvalid, // no tests
```
jak-pan, line 153: still true?

pallets/xyk/src/lib.rs
enthusiastmartin: test added

pallets/xyk/src/lib.rs
enthusiastmartin: this is bit tricky to come up with scenario because this error is returns when buy price + fee overflows - so i cant think of way to generate such scenario.

PR #93, opened 2021-07-05
feat(multipayment): use fallback in set_currency

Roznovjak:
## Description
Use fallback account and price in set_currency() if the corresponding pool does not exist.
Add missing tests found by code coverage tools.


pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -371,15 +371,17 @@ impl<T: Config> Pallet<T> {
...
 		let fee = base_fee.saturating_add(adjusted_weight_fee);
 
-		Self::swap_currency(who, fee)?;
-		T::MultiCurrency::withdraw(currency, who, fee)?;
-
-		Ok(())
+        let result = Self::swap(who, fee)?;
+        match result {
+			PaymentSwapResult::Transferred => Ok(()),
+			PaymentSwapResult::Native | PaymentSwapResult::Swapped => T::MultiCurrency::withdraw(CORE_ASSET_ID, who, fee),
```
enthusiastmartin, line 382: We need to check the result of withdraw too. Because if it fails, the tx should fail too.

enthusiastmartin (commented):
looks ok. However there is a change in xyk pallet which has introduced trading limits. So the test would need some adjustments when merged.

pallets/transaction-multi-payment/src/lib.rs
Roznovjak: I believe that the result (`withdraw` method returns `DispatchResult`) is returned from the `match` statement. This test should prove it https://github.com/galacticcouncil/Basilisk-node/blob/076984b3988833b1144b5a397520b3fad21f0b9a/pallets/transaction-multi-payment/src/tests.rs#L142
Or am I wrong?

enthusiastmartin: yes. correct.  I haven;t realized it.

enthusiastmartin (changes requested):
Looks ok.

However the  withdrawal is not yet enable in the runtime. It needs to be set to Yes.

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -371,15 +371,17 @@ impl<T: Config> Pallet<T> {
...
+			PaymentSwapResult::Error => Err(Error::<T>::Overflow.into()),
```
enthusiastmartin, line 383: I would change this error to something more meaningful. I believe this will be reported back to the user. Overflow wont make much sense.

enthusiastmartin (changes requested):

pallets/transaction-multi-payment/src/lib.rs
Roznovjak: I have decided to remove the `PaymentSwapResult::Error` enum variant because it's unused.

enthusiastmartin: ok.makes sense

enthusiastmartin (approved):

Issue #75, opened 2021-06-28
Companion for HydraDX-273


jak-pan:
We should finish https://github.com/galacticcouncil/HydraDX-node/issues/273 before stabilizing the chain.

PR #95, opened 2021-07-09
feat: access amm constants from exchange api

Roznovjak:
## Description
Make AMM constants( min_trading_limit, min_pool_liquidity, max_in_ratio, max_out_ratio ) accessible from exchange public API

## Related Issue
Fixes: #75


primitives/src/traits.rs
```diff
@@ -96,6 +97,22 @@ pub trait AMM<AccountId, AssetId, AssetPair, Amount> {
...
+		Zero::zero()
+	}
+
+	fn get_max_in_ratio() -> u128 {
+		Zero::zero()
+	}
+
+	fn get_max_out_ratio() -> u128 {
+		Zero::zero()
+	}
```
enthusiastmartin, line 115: Are the default implementations really necessary ?  I would think that this could lead to some issues.

Roznovjak: The default implementations have been removed.

pallets/exchange/src/lib.rs
```diff
@@ -207,6 +207,30 @@ pub mod pallet {
 	pub type ExchangeAssetsIntentions<T: Config> =
 		StorageMap<_, Blake2_128Concat, (AssetId, AssetId), Vec<Intention<T>>, ValueQuery>;
 
+	#[allow(dead_code)]
```
jak-pan, line 210: Do we need this here if we don't use it anywhere?

Roznovjak: clippy complains that the functions we have added are not used anywhere.

jak-pan: do we need the functions tho?

jak-pan: Ah is this what creates the constants? We should add it to xyk (since it's limit mostly of xyk not exchange) too if yes. Or maybe I'm too fast and missed something.

Roznovjak: I would say yes because the task was to make AMM constants( min_trading_limit, min_pool_liquidity, max_in_ratio, max_out_ratio ) accessible from exchange public API.

Roznovjak: The constants are also accessible from xyk and lbp pallet.

jak-pan: It makes sense from the UI standpoint to have it here but it might be the case that different AMM implementations have different limits. What do we do in that case? Use the strict one for everything? We'll need to document this properly in such case.

Roznovjak: Yes, you are right. Different AMMs use different values. But exchange pallet uses only one pool, specified in Config::AMMPool. So we always get what we expect, the values used by exchange pallet.

jak-pan: Ok, we'll want it to use multiple pools later but that will probably be task for future then.

jak-pan (approved):

PR #96, opened 2021-07-09
fix(xyk): create_pool respects min liquidity limit

Roznovjak:
## Description
Create pool only if both amounts are greater than `MIN_POOL_LIQUIDITY`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -511,7 +511,7 @@ dependencies = [
 
 [[package]]
 name = "basilisk-runtime"
-version = "10.0.0"
+version = "11.0.0"
 dependencies = [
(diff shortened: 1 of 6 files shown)


enthusiastmartin (commented):
Yes, This makes sense. 

However, I am still not sure that we need to bump runtime versions every time, even for such small change which does not have impact on api or storage.

pallet version ok but runtime ?

@jak-pan what do you think ?

jak-pan:
@enthusiastmartin We need to bump runtime / spec version because it would panic otherwise. One node would accept this as valid and the other one would not.

jak-pan (approved):

PR #99, opened 2021-07-11
chore: use latest math package

enthusiastmartin:


Roznovjak (commented):
I think that the runtime version needs to be bumped.

enthusiastmartin:
> I think that the runtime version needs to be bumped.

yes, that's the thing i have problem with. Previous merged PR bumped the runtime version but that was not released yet. So the same runtime version can be used with this change too. 

@jak-pan am i right ? 

I know that this is something we already talked about but i am still not happy bumping runtime version with every PR.

jak-pan:
Always bump the runtime if needed, it's not because of release but because the developers that will be running this and our own infrastructure that will be testing compatiility between commits. If you don't bump it someone might run this successfully for a while and then it will crash for him because the runtime will be incompatible at given function call. We want to avoid these errors.

PR #100, opened 2021-07-12
ci: restrict gh workflow

enthusiastmartin:
Added change to run gh workflows only if there is a changes in any rust file.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -6,6 +6,8 @@ on:
       - master
   workflow_dispatch:
   pull_request:
+    paths:
+      - '**.rs'
 
 jobs:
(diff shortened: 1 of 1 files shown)


lumir-mrkva:
@enthusiastmartin i would also add `toml`, `json` and `yaml`

enthusiastmartin:
> @enthusiastmartin i would also add `toml`, `json` and `yml`

Sure, we can tune it.

I wanted to try it out first. But the only thing is that the result of build is needed here ( hence the "some checks are not completed").

I was wondering how to resolve it. With jakub, we talked about a possibility to let it run but skip the steps somehow (build,test etc) and it would successfully pass.

I haven't look at that yet.

PR #104, opened 2021-07-15
feat: duster pallet

enthusiastmartin:
Added duster and offchain duster pallets.


pallets/duster/src/lib.rs
```diff
@@ -0,0 +1,263 @@
...
+				// Not sure if we should fail here?
+				let _ = frame_system::Pallet::<T>::dec_providers(&account);
+			}
+
+			Self::deposit_event(Event::Dusted(account, dust));
+
+			// Ignore the result, it fails - no problem.
+			let _ = Self::reward_duster(&who, currency_id, dust);
+
+			Ok(().into())
```
apopiak, line 202: You might want to refund the fees if the dusting was successful. That way you can maybe get away with a smaller reward.

lumir-mrkva (approved):

Issue #106, opened 2021-07-15
Implement on-chain oracles


Roznovjak:
Implement on-chain oracles providing price estimates for asset pairs based on previous trading data.
Aggregate data of different length periods( e.g. 1, 10, 100 and 1000 blocks ) and calculate average prices.
Values tracked: trading prices and volumes, total pool liquidity.

PR #112, opened 2021-07-22
feat: price-oracle pallet

Roznovjak:
## Description
Implement on-chain oracles providing price estimates for asset pairs based on previous trading data.
Aggregate data of different length periods( e.g. 1, 10, 100 and 1000 blocks ) and calculate average prices.
Values tracked: trading prices and volumes, total pool liquidity.

## Related Issue
Fixes: #106


pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,377 @@
...
+	fn index_mut(&mut self, index: usize) -> &mut Self::Output {
+		&mut self.bucket[index]
+	}
+}
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(RuntimeDebug, Encode, Decode, Copy, Clone, PartialEq, Eq, Default)]
+pub struct PriceEntry {
+	price: Price,
+	amount: Balance,
```
jak-pan, line 179: trade_amount probably? (less confusion)

Roznovjak:
Our use case is to allow another pallets to call `on_trade()` handler and populate the `PriceBuffer` with trading data. At the end (or at the beginning) of a block, this pallet processes the data from the price buffer. There is no upper limit on how many entries the buffer can have. If the buffer contains multiple trades of the same asset, we calculate average value.
We need to figure out how to benchmark `on_trade()` method and whether to process the data from the `PriceBuffer` in `on_initialize` or in `on_finalize`. It should be ok to update prices in the next block ( in the case of `on_initialize`). 

@jak-pan @lumir-mrkva @apopiak your expertise would be greatly appreciated.

pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+	#[pallet::event]
+	pub enum Event<T: Config> {}
+
+	#[pallet::storage]
+	#[pallet::getter(fn price_buffer)]
+	pub type PriceBuffer<T: Config> = StorageMap<_, Blake2_128Concat, AssetPairId, Vec<PriceEntry>, ValueQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn price_data_ten)]
+	pub type PriceDataTen<T: Config> = StorageValue<_, Vec<(AssetPairId, BucketQueue)>, ValueQuery>;
```
green-jay, line 81: docs could be helpful here

green-jay:
polkadotJS types definition seems to be missing

pallets/price-oracle/src/types.rs
```diff
@@ -0,0 +1,215 @@
...
+
+#[cfg(feature = "std")]
+use serde::{Deserialize, Serialize};
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(RuntimeDebug, Encode, Decode, Copy, Clone, PartialEq, Eq, Default)]
+pub struct PriceEntry {
+	pub price: Price,
+	pub amount: Balance,
+	pub liq_amount: Balance,
```
green-jay, line 34: I would prefer complete words in variable names when possible (as liq could be ambiguous)

pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+	pub trait Config: frame_system::Config {
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// Weight information for the extrinsics.
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::error]
+	pub enum Error<T> {
+		PriceComputationError,
```
green-jay, line 68: error docs

pallets/price-oracle/src/types.rs
```diff
@@ -0,0 +1,215 @@
...
+		)
+	}
+}
+
+impl PriceInfoCalculation<PriceInfo> for PriceInfo {
+	fn calculate_price_info(entries: &[PriceEntry]) -> Option<PriceInfo> {
+		let intermediate_result: Vec<PriceEntry> = entries
+			.iter()
+			.map(|x| PriceEntry {
+				price: x.price.mul(Price::from(x.liq_amount)),
```
green-jay, line 143: type safe mul might be better? e.g. checked or saturated

pallets/price-oracle/src/types.rs
```diff
@@ -0,0 +1,215 @@
...
+
+#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+#[derive(RuntimeDebug, Encode, Decode, Copy, Clone, PartialEq, Eq)]
+pub struct BucketQueue {
+	bucket: Bucket,
+	last: u32,
+}
+
+impl BucketQueue {
+	// make sure that BUCKET_SIZE != 0
```
green-jay, line 166: did you? :)

pallets/price-oracle/src/types.rs
```diff
@@ -0,0 +1,215 @@
...
+				price: Price::zero(),
+				amount: Balance::zero(),
+				liq_amount: Balance::zero(),
+			},
+			|a, b| &a + b,
+		)
+	}
+}
+
+pub const BUCKET_SIZE: u32 = 10;
```
green-jay, line 89: This is probably going to be configured from time to time so might be better as a pallet const
Maybe NonZeroU32 type can be used instead?

green-jay:
> Our use case is to allow another pallets to call `on_trade()` handler and populate the `PriceBuffer` with trading data. At the end (or at the beginning) of a block, this pallet processes the data from the price buffer. There is no upper limit on how many entries the buffer can have. If the buffer contains multiple trades of the same asset, we calculate average value.
> We need to figure out how to benchmark `on_trade()` method and whether to process the data from the `PriceBuffer` in `on_initialize` or in `on_finalize`. It should be ok to update prices in the next block ( in the case of `on_initialize`).
> 
> @jak-pan @lumir-mrkva @apopiak your expertise would be greatly appreciated.

afaik rule of thumb is keep the load on on_finalize minimized, so all the heavier operations should always be processed in on_inialize

pallets/price-oracle/src/types.rs
```diff
@@ -0,0 +1,215 @@
...
+
+	fn get_last(&self) -> PriceInfo {
+		self.bucket[self.last as usize]
+	}
+
+	fn calculate_average(&self) -> PriceInfo {
+		let sum = self.bucket.iter().sum::<PriceInfo>();
+		PriceInfo {
+			avg_price: sum.avg_price.checked_div(&Price::from(Self::BUCKET_SIZE as u128)).expect("avg_price is valid value; BUCKET_SIZE is non-zero integer; qed"),
+			volume: sum.volume.checked_div(Self::BUCKET_SIZE as u128).expect("avg_price is valid value; BUCKET_SIZE is non-zero integer; qed"),
```
green-jay, line 199: also unsigned

jak-pan:
> > Our use case is to allow another pallets to call `on_trade()` handler and populate the `PriceBuffer` with trading data. At the end (or at the beginning) of a block, this pallet processes the data from the price buffer. There is no upper limit on how many entries the buffer can have. If the buffer contains multiple trades of the same asset, we calculate average value.
> > We need to figure out how to benchmark `on_trade()` method and whether to process the data from the `PriceBuffer` in `on_initialize` or in `on_finalize`. It should be ok to update prices in the next block ( in the case of `on_initialize`).
> > @jak-pan @lumir-mrkva @apopiak your expertise would be greatly appreciated.
> 
> afaik rule of thumb is keep the load on on_finalize minimized, so all the heavier operations should always be processed in on_inialize

Did this change? We were told the exact opposite before when we were doing exchange. The only thing you need to make sure is to reserve weight in on-initialize -> That's why the exchange order matching is done in on-finalize. (The data structures can be still cleaned for that block so we don't need to save unnecessary data to blockchain)

apopiak:
The most important thing is to make sure that the weight used is predictable. As long as all weight consumed in the hooks is tracked it does not matter whether you execute in `on_finalize`, `on_initialize` or `on_idle` from that perspective.
So a thing to avoid is to execute arbitrary logic in `on_finalize` e.g. iterating a storage map whose number of keys you don't know.
If you do something where you don't know the weight ahead of time, you're better of doing it in `on_initialize` and keeping track of the weight as you're executing (and limiting it to the total block weight). If you can estimate/limit the weight ahead of time you can do it in `on_finalize`.
As Jakub mentioned there are, of course, also other considerations why you might choose one over the other.

apopiak:
> There is no upper limit on how many entries the buffer can have.

This line makes me uncomfortable xD
You might want to 
a) argue why it's still safe (even though it might lead to unbounded execution time)
or b) change that

pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,201 @@
...
+		let mut num_of_processed_buckets = 0u32;
+		let mut num_of_processed_trades = 0u32;
+
+		PriceDataTen::<T>::mutate(|data_ten| -> DispatchResult {
+			for (asset_pair_id, data) in data_ten.iter_mut() {
+				let maybe_price = <PriceBuffer<T>>::try_get(asset_pair_id);
+				let result = if let Ok(prices) = maybe_price {
+					num_of_processed_buckets.checked_add(1).ok_or(Error::<T>::UpdateDataOverflow)?;
+					num_of_processed_trades.checked_add(prices.len().try_into().map_err(|_| Error::<T>::UpdateDataOverflow)?).ok_or(Error::<T>::UpdateDataOverflow)?;
+					PriceInfo::calculate_price_info(prices.as_slice()).ok_or(Error::<T>::PriceComputationError)?
```
apopiak, line 128: This looks quadratic to me (number of asset pairs * number of prices per pair). You thus need to make sure to cover the max value of each in the benchmark to get a sensible linear estimate. (The linear regression run on the benchmark data thinks the number of asset pairs and the number of prices are independent, but they are not.)

pallets/price-oracle/src/types.rs
Roznovjak: We can't use NonZeroU32 because its constructor is not `const`. The same problem applies also for pallet config parameters.

pallets/price-oracle/src/types.rs
Roznovjak: not sure what you mean

Roznovjak:
> > There is no upper limit on how many entries the buffer can have.
> 
> This line makes me uncomfortable xD
> You might want to
> a) argue why it's still safe (even though it might lead to unbounded execution time)
> or b) change that

Sorry for making you feel uncomfortable. :) The number of entries in the price buffer is the same as the number of trades performed in the last block, and this is limited by the number of buy/sell TXs we can fit into a block. I also assume (it needs to be verified by us) that a trading TX is more computationally expensive compared to the price calculation we perform in this pallet. It may have the following side effect: the more trading TXs we have in a block, the less number of TXs we can have in the next one. If the previous assumptions are true, I think it may be safe to use the approach we already use.

pallets/price-oracle/src/types.rs
green-jay: total nitpick :) BUCKET_SIZE is a non-zero *unsigned* integer or simply put - a positive integer

apopiak:
Ah, you plan on including the `on_finalize` overhead as part of the transaction causing that overhead, right?

pallets/price-oracle/src/lib.rs
jak-pan: Hmm. If we store intermediate value after each trade with total weight we shouldn't need to recalculate it. Like so -> If you look at the equations you only need total weight and previous price to calculate current price instead of doing recalc of everything each time. https://docs.google.com/spreadsheets/d/1hYP8Y0qDbKOFSHcnP0Q3oyJmOwXTDLV79851Be5Hrgo/edit?usp=sharing

Roznovjak: @jak-pan it makes sense. It may help to solve the problem with on_trade() where it was complicated to calculate the weight because we don't know the number of already executed trades stored in the price buffer. I am going to test it.

pallets/price-oracle/src/types.rs
Roznovjak: Any idea how this can be done?

pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+impl<T: Config> Pallet<T> {
+	pub fn on_create_pool(asset_pair: AssetPair) {
+		let data = PriceDataTen::<T>::get();
+		if !data.iter().any(|bucket_tuple| bucket_tuple.0 == asset_pair.name()) {
+			let incremented_asset_count = if let Some(count) = Self::num_of_assets().checked_add(1) {
+				count
+			} else {
+				// We don't want to throw an error here because this method is used in different extrinsics.
+				// We also do not expect to have more than 2^32 assets registered.
+				return
```
enthusiastmartin, line 136: I don't think that just returning here is correct. I believe you can panic here. 

Because if you reach this state - although unlikely - the stuff will not work properly anymore and you'd fail silently.

pallets/xyk/src/mock.rs
```diff
@@ -145,6 +145,7 @@ impl Config for Test {
 	type NativeAssetId = NativeAssetId;
 	type WeightInfo = ();
 	type GetExchangeFee = ExchangeFeeRate;
+	type AMMHandler = ();
```
enthusiastmartin, line 148: I'd consider splitting into separate handlers : OnCreatePool, OnTrade. 

Reason being that in future we'd like to make this pallet more generic and there can be a case where only one of them is needed.

primitives/src/traits.rs
```diff
@@ -99,10 +126,34 @@ pub trait AMM<AccountId, AssetId, AssetPair, Amount> {
...
+
+impl<AccountId, AssetId, AssetPair, Balance> AMMHandlers<AccountId, AssetId, AssetPair, Balance> for () {
+	fn on_create_pool(_asset_pair: AssetPair) {}
+	fn on_trade(_amm_transfer: &AMMTransfer<AccountId, AssetId, AssetPair, Balance>, _liq_amount: Balance) {}
+	fn on_trade_weight() -> Weight {
+		Weight::zero()
+	}
+}
+
+pub trait AssetPairAccountIdFor<AssetId: Sized, AccountId: Sized> {
```
enthusiastmartin, line 157: just a a note : I created an issue  regarding this #129.  This can be a start.

primitives/src/traits.rs
```diff
@@ -99,10 +126,34 @@ pub trait AMM<AccountId, AssetId, AssetPair, Amount> {
...
+	/// Include a trade in the average price calculation of the price-oracle pallet.
+	fn on_trade(amm_transfer: &AMMTransfer<AccountId, AssetId, AssetPair, Balance>, liq_amount: Balance);
+	/// Known `on_trade` overhead. Needs to be specified here if we don't want to make AMM pools tightly coupled with the price oracle pallet.
+	fn on_trade_weight() -> Weight;
+}
+
+impl<AccountId, AssetId, AssetPair, Balance> AMMHandlers<AccountId, AssetId, AssetPair, Balance> for () {
+	fn on_create_pool(_asset_pair: AssetPair) {}
+	fn on_trade(_amm_transfer: &AMMTransfer<AccountId, AssetId, AssetPair, Balance>, _liq_amount: Balance) {}
+	fn on_trade_weight() -> Weight {
```
enthusiastmartin, line 152: Better descriptions here.. I still can't tell when the `on_trade_weight` is actually called and what is supposed to do.

pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// Pool was registered. [AssetPair]
+		PoolRegistered(AssetPair),
+	}
+
+	/// The number of assets registered and handled by this pallet.
+	#[pallet::storage]
+	#[pallet::getter(fn num_of_assets)]
+	pub type NumOfTrackedAssets<T: Config> = StorageValue<_, u32, ValueQuery>;
```
enthusiastmartin, line 84: `TrackedAssetsCount` . maybe?

And it can be AssetId type instead of u32

pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+
+pub mod weights;
+use weights::WeightInfo;
+
+mod benchmarking; // TODO: rebenchmark
+
+// Re-export pallet items so that they can be accessed from the crate namespace.
+pub use pallet::*;
+
+pub type AssetPairId = Vec<u8>;
```
enthusiastmartin, line 47: This type looks bit strange to me. 

Perhaps it just the name.


A description would be good. I can;t tell what it should represent.

enthusiastmartin (commented):
Few more comments. I haven't  got through  it all yet.

pallets/price-oracle/src/lib.rs
Roznovjak: We use `u32` because we use the value here https://github.com/galacticcouncil/Basilisk-node/blob/534465213edee32e2a9ad24466bd1ebf1685b95a/pallets/price-oracle/src/lib.rs#L119 and autogenerated parametrized weight functions uses `u32`https://github.com/galacticcouncil/Basilisk-node/blob/534465213edee32e2a9ad24466bd1ebf1685b95a/pallets/price-oracle/src/weights.rs#L82

pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+	#[pallet::getter(fn num_of_assets)]
+	pub type NumOfTrackedAssets<T: Config> = StorageValue<_, u32, ValueQuery>;
+
+	/// Processed or partially processed data generated by trades.
+	/// Data generated by trades are processed sequentially.
+	/// Each new entry is combined with the previous value to produce new intermediate value.
+	/// The last entry creates the resulting average price and volume.
+	#[pallet::storage]
+	#[pallet::getter(fn price_accumulator)]
+	pub type PriceDataAccumulator<T: Config> = StorageMap<_, Blake2_128Concat, AssetPairId, PriceEntry, ValueQuery>;
```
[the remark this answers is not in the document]
Roznovjak: Maybe it would be even better to use `Identity` hasher with 32 bit keys (`AssetPairId` is u32 ).

pallets/price-oracle/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {}
+}
+
+impl<T: Config> Pallet<T> {
+	pub fn on_create_pool(asset_pair: AssetPair) {
+		let data = PriceDataTen::<T>::get();
+		if !data.iter().any(|bucket_tuple| bucket_tuple.0 == asset_pair.name()) {
```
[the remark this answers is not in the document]
Roznovjak: Yes it is necessary. We need to check the existence of an asset pair and increase the asset_count if the asset pair exists because it is used in weigh calculation here https://github.com/galacticcouncil/Basilisk-node/blob/534465213edee32e2a9ad24466bd1ebf1685b95a/pallets/price-oracle/src/weights.rs#L82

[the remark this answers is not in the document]
enthusiastmartin: yes, i know it is necessary now. I really meant if this could not be  done in a way where there is no need to iterate over it just to find the pair

pallets/price-oracle/src/lib.rs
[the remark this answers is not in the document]
enthusiastmartin: Identity no. As far as i know - it should be used only when the key is already a hash.

pallets/price-oracle/src/lib.rs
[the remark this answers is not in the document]
Roznovjak: I'm not aware of a better solution. We need to use an iteration because we use a vector. The reason behind this is that we need to iterate over all values at the end of each block, and because it is advised to use vectors over maps (faster iterations). https://substrate.dev/docs/en/knowledgebase/runtime/storage#iterating-over-storage-maps

pallets/price-oracle/src/mock.rs
```diff
@@ -0,0 +1,196 @@
...
+	type NativeAssetId = NativeAssetId;
+	type WeightInfo = ();
+	type GetExchangeFee = ExchangeFeeRate;
+	type AMMHandler = PriceOracleHandler<Test>;
+}
+
+
+parameter_type_with_key! {
+	pub ExistentialDeposits: |_currency_id: AssetId| -> Balance {
+		Zero::zero()
```
enthusiastmartin, line 168: Set existential depoist to something other than 0. Eg. 1_000_000.

pallets/price-oracle/src/types.rs
```diff
@@ -0,0 +1,222 @@
...
+	fn sum<I>(iter: I) -> Self
+	where
+		I: Iterator<Item = &'a Self>,
+	{
+		iter.fold(
+			PriceEntry {
+				price: Price::zero(),
+				trade_amount: Balance::zero(),
+				liquidity_amount: Balance::zero(),
+			},
```
enthusiastmartin, line 84:
```suggestion
			PriceEntry::zero(),
```

or if default was implemented: 

```suggestion
			PriceEntry::default(),
```

pallets/price-oracle/src/types.rs
```diff
@@ -0,0 +1,222 @@
...
+impl<'a> Sum<&'a Self> for PriceInfo {
+	fn sum<I>(iter: I) -> Self
+	where
+		I: Iterator<Item = &'a Self>,
+	{
+		iter.fold(
+			PriceInfo {
+				avg_price: Price::zero(),
+				volume: Balance::zero(),
+			},
```
enthusiastmartin, line 149:
```suggestion
			PriceInfo::default(),
```

pallets/price-oracle/src/lib.rs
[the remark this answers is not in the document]
lumir-mrkva: why not to just use a map where this check could be O(1) ?

PR #113, opened 2021-07-25
Review Basilisk (SBP Milestone 3)

apopiak:
Review of the Basilisk runtime and included pallets for the Substrate Builders' Program Milestone 3.

Below is a summary of what I consider the most important points:

#### Runtime
+ make sure to generate weights
+ what's your dust management strategy?
+ you seem to have copied the Hydra pallets which IMO adds maintenance burden

#### pallet-asset-registry
+ don't use `unwrap` in the runtime
+ suggestion: check the genesis config for sanity

#### pallet-exchange
+ I don't think you cover the worst case for the `on_finalize` weight

#### pallet-lbp
+ some confusion about naming/semantics of pausing and pool activity

#### pallet-nft
+ `mint` is not marked `#[transactional]` but should be
+ benchmarking does not assume worst case

#### pallet-transaction-multi-payment
+ benchmarking does not assume worst case

### DIFF
--- a/pallets/asset-registry/src/lib.rs
+++ b/pallets/asset-registry/src/lib.rs
@@ -91,9 +91,16 @@ pub mod pallet {
 	#[pallet::genesis_build]
 	impl<T: Config> GenesisBuild<T> for GenesisConfig<T> {
 		fn build(&self) {
+			// REVIEW: You probably want to make sure the genesis config is sane?
+			assert_ne!(self.core_asset_id, self.next_asset_id, "core asset id needs to be unique");
+			// REVIEW: maybe even:
+			assert!(self.core_asset_id < self.next_asset_id, "core asset id needs to be unique");
 			CoreAssetId::<T>::put(self.core_asset_id);
 			NextAssetId::<T>::put(self.next_asset_id);
+			// REVIEW: Maybe track them in a btree-set? Might be overkill, not sure.
 			self.asset_ids.iter().for_each(|(name, asset_id)| {
+				// REVIEW: maybe:
+				assert!(asset_id < &self.next_asset_id, "asset ids need to be unique");
 				AssetIds::<T>::insert(name, Some(asset_id));
 			})
 		}
@@ -102,9 +109,12 @@ pub mod pallet {
 
 impl<T: Config> Pallet<T> {
 	/// Create asset for given name or return existing AssetId if such asset already exists.
+	/// 
+	/// Note: Does not check that `name` is of a sane length, length bound checks fall to the caller.
 	pub fn get_or_create_asset(name: Vec<u8>) -> Result<T::AssetId, DispatchError> {
 		if <AssetIds<T>>::contains_key(&name) {
-			Ok(<AssetIds<T>>::get(&name).unwrap())
+			// REVIEW: Add a "proof"
+			Ok(<AssetIds<T>>::get(&name).expect("checked existence of value; qed"))
 		} else {
 			let asset_id = Self::next_asset_id();
 			let next_id = asset_id.checked_add(&One::one()).ok_or(Error::<T>::NoIdAvailable)?;
--- a/pallets/exchange/benchmarking/src/lib.rs
+++ b/pallets/exchange/benchmarking/src/lib.rs
@@ -294,14 +294,18 @@ benchmarks! {
 	on_finalize_for_one_sell_extrinsic {
 		let creator = funded_account::<T>("creator", 100);
 		let seller = funded_account::<T>("seller", 101);
-
+		
 		let asset_a: AssetId = 1;
 		let asset_b: AssetId = 2;
 		let amount : Balance = 10_000_000_000;
 		let discount = false;
-
+		
 		initialize_pool::<T>(creator, asset_a, asset_b, amount, Price::from(1))?;
-
+		
+		// REVIEW: I don't think one sell covers the worst-case on_finalize overhead.
+		// I would expect `resolve_matched_intentions` not to be called with just one intention
+		// because it's based on the matching. The on_finalize overhead of one sell should probably
+		// be based on max-buy-intentions already in the queue and vice-versa for one buy.
 		pallet_exchange::Pallet::<T>::sell(
 			RawOrigin::Signed(seller.clone()).into(),
 			asset_a,
--- a/pallets/exchange/src/lib.rs
+++ b/pallets/exchange/src/lib.rs
@@ -375,8 +375,10 @@ impl<T: Config> Pallet<T> {
 		b_in_intentions: &[Intention<T>],
 	) {
 		let mut b_copy = b_in_intentions.to_owned();
+		// REVIEW: You can sort `mut` slices, so you could pass a mut slice and avoid a clone here.
 		let mut a_copy = a_in_intentions.to_owned();
 
+		// REVIEW: nitpick: change `a, b` to something else like `l, r` to avoid naming collision.
 		b_copy.sort_by(|a, b| b.amount_in.cmp(&a.amount_in));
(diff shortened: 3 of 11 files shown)


pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -218,7 +218,8 @@ pub mod pallet {
 		/// When currency is set, fixed fee is withdrawn from the account to pay for the currency change
 		///
 		/// Emits `CurrencySet` event when successful.
-		#[pallet::weight((<T as Config>::WeightInfo::set_currency(), DispatchClass::Normal, Pays::No))]
+		// REVIEW: You would want to charge before the transaction is executed, no?
```
enthusiastmartin, line 221: This would be preferable way. But here we need to pay the fee in the currency which is being set with this call.  

That was the reason to manually withdraw a fee within the call. 

Any tip if there is easy to pay the fee before the is executed but using the selected currency ?

jak-pan: We should have a filter on this call and it should get rejected by a node if you don't have enough to pay for the fee after the fact so spam should not be possible (only on rpc side which is reasonable I think)

jak-pan:
> Review of the Basilisk runtime and included pallets for the Substrate Builders' Program Milestone 3.
> 
> Below is a summary of what I consider the most important points:
> 
> #### Runtime
> * make sure to generate weights
In progress

> * what's your dust management strategy?
We are developing pallet-duster https://github.com/galacticcouncil/HydraDX-node/pull/288 please feel free to comment / review on it @enthusiastmartin if ready.

> * you seem to have copied the Hydra pallets which IMO adds maintenance burden
We did this because basilisk is being audited and we'll be removing them from HydraDX soon.

Others 👍🏻 on it.

runtime/src/lib.rs
```diff
@@ -446,6 +450,7 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 	type OnValidationData = ();
 	type SelfParaId = ParachainInfo;
 	type OutboundXcmpMessageSource = ();
+	// REVIEW: I'd understand holding off on XCMP, but you're not even handling DMPs, why?
```
jak-pan, line 453: It's wip #114

runtime/src/lib.rs
```diff
@@ -589,12 +596,14 @@ impl pallet_elections_phragmen::Config for Runtime {
...
 	type DesiredRunnersUp = DesiredRunnersUp;
 	type TermDuration = TermDuration;
+	// REVIEW: Same note here about benchmarks.
 	type WeightInfo = ();
 }
 
 parameter_types! {
 	pub const CouncilMotionDuration: BlockNumber = 5 * DAYS;
 	pub const CouncilMaxProposals: u32 = 20;
+	// REVIEW: What's the use of having a council of one?
```
jak-pan, line 606: We'll be increasing it after the chain runs.

pallets/lbp/src/lib.rs
```diff
@@ -504,7 +505,9 @@ pub mod pallet {
 
 				ensure!(pool.paused, Error::<T>::PoolIsNotPaused);
 
-				ensure!(Self::is_after_sale(&pool), Error::<T>::CannotUnpauseEndedPool);
+				// REVIEW: You want to ensure the pool is running, no? Or do you want to allow
```
jak-pan, line 508: we want to allow paused start -> docs?

apopiak: sure, but the function did sth else AFAICT

pallets/transaction-multi-payment/src/lib.rs
apopiak: You could probably introduce a SignedExtension that inspects the call and checks the balance for the configured currency, yeah.

pallets/asset-registry/src/lib.rs
```diff
@@ -91,9 +91,16 @@ pub mod pallet {
...
+			// REVIEW: You probably want to make sure the genesis config is sane?
+			assert_ne!(self.core_asset_id, self.next_asset_id, "core asset id needs to be unique");
+			// REVIEW: maybe even:
+			assert!(self.core_asset_id < self.next_asset_id, "core asset id needs to be unique");
 			CoreAssetId::<T>::put(self.core_asset_id);
 			NextAssetId::<T>::put(self.next_asset_id);
+			// REVIEW: Maybe track them in a btree-set? Might be overkill, not sure.
 			self.asset_ids.iter().for_each(|(name, asset_id)| {
+				// REVIEW: maybe:
+				assert!(asset_id < &self.next_asset_id, "asset ids need to be unique");
```
enthusiastmartin, line 103: Yes, this pallet had the refactor coming for quite some time.

And now, we're completely reworking our asset-registry. It will include better handling of assets in  genesis config too.

runtime/Cargo.toml
```diff
@@ -25,6 +25,8 @@ trie-db = { version = "0.22.0", default-features = false }
 smallvec = "1.6.1"
 
 # local dependencies
+# REVIEW: Seems like you copied the pallets from HydraDX node. Seems like a bad idea as it increases
+#   the maintenance burden.
```
enthusiastmartin, line 29: This has been changed in HydraDX node repo. And now, the hydra repo uses pallets from Basilisk to remove the maintenance burden.

runtime/src/lib.rs
```diff
@@ -312,6 +312,7 @@ impl pallet_timestamp::Config for Runtime {
 }
 
 parameter_types! {
+	// REVIEW: How do you avoid dust?
```
enthusiastmartin, line 315: Dust handling is done by pallet-duster and pallet-offchain-duster.

runtime/src/lib.rs
```diff
@@ -373,6 +376,7 @@ impl pallet_sudo::Config for Runtime {
 }
 
 parameter_type_with_key! {
+	// REVIEW: How do you plan on handling dust?
```
enthusiastmartin, line 379: Dust handling is done by pallet-duster and pallet-offchain-duster.

pallets/exchange/src/lib.rs
```diff
@@ -390,9 +392,13 @@ impl<T: Config> Pallet<T> {
 			let mut bvec = Vec::<Intention<T>>::new();
 			let mut total = 0;
 
+			// REVIEW: I think you can avoid the allocation of the vec here if you use `take_while`
+			// https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.take_while
+			// on an iterator of the slice.
```
enthusiastmartin, line 397: Thanks. 
I finally refactored the matching part of the algorithm - using also your take_while suggestion. Also removed the unnecessary vec copy.

[PR Here](https://github.com/galacticcouncil/Basilisk-node/blob/cf67165c6eca7fdccc1b48f630ed4542d70849f9/pallets/exchange/src/lib.rs#L389)

pallets/exchange/benchmarking/src/lib.rs
```diff
@@ -294,14 +294,18 @@ benchmarks! {
...
 		let discount = false;
-
+		
 		initialize_pool::<T>(creator, asset_a, asset_b, amount, Price::from(1))?;
-
+		
+		// REVIEW: I don't think one sell covers the worst-case on_finalize overhead.
+		// I would expect `resolve_matched_intentions` not to be called with just one intention
+		// because it's based on the matching. The on_finalize overhead of one sell should probably
+		// be based on max-buy-intentions already in the queue and vice-versa for one buy.
```
enthusiastmartin, line 308: we will be refactoring the algorithm in very near future so that would be updated too. so leaving as is for now.

enthusiastmartin:
All review comments have been addressed and either fixed/changed or issue has been created to address it later.

Issue #117, opened 2021-07-29
Bug: README types JSON is invalid & incorrect path


chrisk700:
### Description
The types JSON in the README is not valid and their is a path that needs updating.

### Expected Behavior

1. JSON to be valid
![image](https://user-images.githubusercontent.com/74745915/127468341-780c7fb2-815d-437a-b8d2-8232fa19f8f0.png)
2. Directory to be found, polkadot-launch to start
![image](https://user-images.githubusercontent.com/74745915/127470673-2841759e-b626-4f4c-a93f-570e724cc0e4.png)

### Actual Behavior

1. JSON is invalid
![image](https://user-images.githubusercontent.com/74745915/127468386-0c429d84-3f27-4544-9708-e92a04240445.png)

2. Path not found
![image](https://user-images.githubusercontent.com/74745915/127470585-92fc2c95-4fb2-48bd-bd51-7c5758bf9a84.png)

### Possible Fix

PR to be submitted

### Steps to Reproduce

1. Follow steps in README.md
2. Observe error

### Context

Unable to complete the initial onboarding

### Your Environment

n/a

jak-pan:
We're aware of this, unfortunately polkadot-js/apps don't allow using alias which is required for everything to work properly in js/apps. And it looks like nobody else cares for now. @lumir-mrkva do you think we can do something else so that our dev-chain is properly recognized as HydraDX and use our provided types automatically? I see no other option unless this is fixed.

https://github.com/polkadot-js/apps/issues/5835

jak-pan:
Ok looks like the JSON itself was broken. Thanks for the headsup!

PR #118, opened 2021-07-29
fix: Onboarding JSON types and path fixes [skip ci]

chrisk700:
## Description
As part of testing the README.md onboarding an issue was found with the types JSON being invalid and an incorrect path in one of the commands.

## Related Issue

Fixes: #117 

## Motivation and Context

n/a

## How Has This Been Tested?

n/a

## Checklist:


README.md
```diff
@@ -47,7 +47,7 @@ npm link
 Starts local testnet with 4 relay chain validators and Basilisk as parachain
 
 ```
-cd rococo-local
+cd ../rococo-local
```
lumir-mrkva, line 50: rococo local is located in repo as subdirectory
```suggestion
cd rococo-local
```

lumir-mrkva (changes requested):

README.md
chrisk700: Sure. The issue comes from the section above where we move to the polkadot-local dir where the rococo-local dir does not exist.

PR #132, opened 2021-08-22
Fix runner image

fakirAyoub:
Fix Runner image ID


enthusiastmartin:
I don't understand why 142 files has changed in this PR. It seems that only one file should have changed. 

@fakirAyoub  Perhaps it would be better to re-do it, so we avoid .... something ...

fakirAyoub:
Yes, weird. I think the best would be to change the AMI ID directly in the master branch, wdyt? @enthusiastmartin

enthusiastmartin:
Re-done in #134  ( due to changes in many files in this PR ).

Issue #122, opened 2021-08-03
change: duster -  refund fee -> smaller reward


enthusiastmartin:
You might want to refund the fees if the dusting was successful. That way you can maybe get away with a smaller reward.

_Originally posted by @apopiak in https://github.com/galacticcouncil/Basilisk-node/pull/104#discussion_r677274592_

jak-pan:
Deprecated

PR #144, opened 2021-08-29
chore: launch code review

jak-pan:


runtime/src/lib.rs
```diff
@@ -288,6 +289,7 @@ impl pallet_timestamp::Config for Runtime {
 }
 
 parameter_types! {
+	// TODO: Shouldn't this be at least 1?
```
jak-pan, line 292: @enthusiastmartin maybe 1000 like duster?

runtime/src/lib.rs
```diff
@@ -353,6 +355,8 @@ parameter_type_with_key! {
 		// Dusting is handled by duster pallet.
 		// However, to make sure that account is reaped/killed and storage updated, ED must be > 0
 		// On ED = 0 - accounts are never reaped.
+
+		// We might want to set this to the same thing as duster e.g. 1000 for everything as a base
```
jak-pan, line 359: @enthusiastmartin

runtime/src/lib.rs
```diff
@@ -429,6 +434,7 @@ impl pallet_exchange::Config for Runtime {
 }
 
 parameter_types! {
+	// Is this ok?
```
jak-pan, line 437: @enthusiastmartin

enthusiastmartin: the value is ok. I am not sure what the reward amount should be, hence 0 for now.

We can also consider doiing this #122

runtime/src/lib.rs
enthusiastmartin: Yes, we can set all ED to 1000.  

The only is that there is not dust handler - so the dust will be lost at the moment.

runtime/src/lib.rs
enthusiastmartin: Yes, we can set all ED to 1000.  

The only is that there is not dust handler - so the dust will be lost at the moment.

jak-pan: Send to treasury?

runtime/src/lib.rs
jak-pan: The reward should be bigger/same as the amount paid for the tx but smaller than the dust amount.

primitives/src/traits.rs
```diff
@@ -34,7 +34,7 @@ pub struct AMMTransfer<AccountId, AssetId, AssetPair, Balance> {
 
 /// Traits for handling AMM Pool trades.
 pub trait AMM<AccountId, AssetId, AssetPair, Amount> {
-	/// Check if both assets exist in a pool.
+	/// Check if both assets exist in a pool. //TODO: not sure if AssetPair is right here since we don't have asset_in or asset_out when not trading.
```
enthusiastmartin, line 37: Yeah, i was actually thinking that this AMM trait should be more generic to cover various pool implementations.

I have some ideas - but do we want to refactor now ?

runtime/src/currency.rs
```diff
@@ -1,3 +1,5 @@
+// TODO: What is the purpose of this?
```
enthusiastmartin, line 1: No longer used. this was used in original implementation of xcm support.

I will remove it.

primitives/src/traits.rs
jak-pan: Probably not. But let's start working on it?

jak-pan: After launch of crowdloan

runtime/src/lib.rs
jak-pan: Let's actually do 1_000_000 everywhere. It's still 6 decimals so it's like 4 cents if you are dusting bitcoin. Anything else mostly is peanuts. We should do it dynamic tho asap.

primitives/src/traits.rs
enthusiastmartin: #145

runtime/src/lib.rs
jak-pan: Maybe for now add default existential deposit to assetRegistry + set default deposits for generated assets to primitives/MIN_POOL_LIQUIDITY

pallets/asset-registry/src/tests.rs
```diff
@@ -206,3 +208,54 @@ fn set_metadata_works() {
 			);
 		});
 }
+
+#[test]
+fn update_asset() {
```
enthusiastmartin, line 213: These tests actually dont really test the the storage was updated. They only check the result of the update call.

green-jay: added

primitives/src/traits.rs
jak-pan, line 37:
```suggestion
	/// Check if both assets exist in a pool.
```

pallets/lbp/src/lib.rs
```diff
@@ -857,10 +865,17 @@ impl<T: Config> Pallet<T> {
 			let amount_in_with_fee = token_amount_in.checked_add(transfer_fee).ok_or(Error::<T>::Overflow)?;
 
 			ensure!(limit >= amount_in_with_fee, Error::<T>::AssetBalanceLimitExceeded);
-
 			(token_amount_in, amount, assets.asset_in, transfer_fee)
 		};
 
+		// TODO: check if user has enough tokens in?
```
jak-pan, line 871:
```suggestion
```

runtime/src/lib.rs
```diff
@@ -429,30 +428,19 @@ impl pallet_exchange::Config for Runtime {
...
```
jak-pan, line 431:
```suggestion
```

jak-pan (commented):
LGTM

lumir-mrkva (approved):

PR #146, opened 2021-08-30
feat: existential deposits in asset registry

enthusiastmartin:


node/src/chain_spec.rs
```diff
@@ -516,6 +531,7 @@ fn parachain_genesis(
 		asset_registry: AssetRegistryConfig {
 			asset_names: vec![],
 			native_asset_name: TOKEN_SYMBOL.as_bytes().to_vec(),
+			native_existential_deposit: 1_000_000u128,
```
jak-pan, line 534: This should be same as the one in the runtime.

enthusiastmartin: done

Issue #131, opened 2021-08-18
refactor: split primitives crate


enthusiastmartin:
### Description

In Basilisk `primitives` crate there are defined various types used in Basilisk project.

Besides that, there are also various traits to support our pallets.

Therefore, let's say - if you want to use a pallet where a trait has to be implemented - the pallet brings the primitives crates as dependency , although the types can be completely different.

Traits should live in a separate crate - eg. ` basilisk-traits. `
And types used in basilisk in primitives ( as it is now ).

Similar to `orml-traits.`

PR #148, opened 2021-08-31
refactor: split primitives crate

Roznovjak:
## Description
Split `primitives` crate and create new `basilisk-traits` crate.

## Related Issue
Fixes: #131


enthusiastmartin:
Perhaps, we could consider creating a separate repo for basilisk/hydra pallets at this point ? 

And this refactor could be a good start - the traits crate can live there.  It will be needed in Hydra repo as well. In such case - naming should be different. 

@jak-pan  what do you think  ?

jak-pan:
> Perhaps, we could consider creating a separate repo for basilisk/hydra pallets at this point ?
> 
> And this refactor could be a good start - the traits crate can live there. It will be needed in Hydra repo as well. In such case - naming should be different.
> 
> @jak-pan what do you think ?

yes

Roznovjak:
@enthusiastmartin what name do you suggest? What about `hydra-traits`?

enthusiastmartin:
> 
> 
> @enthusiastmartin what name do you suggest? What about `hydra-traits`?

I would also consider creating one repo for all pallets. something lie orml has done for their pallets.

Naming is another question as this should cover both Hydra/Basilisk ( or perhaps some of the pallet could be used in other projects too ).

pallets/asset-registry/Cargo.toml
```diff
@@ -19,6 +19,9 @@ codec = { default-features = false, features = ["derive", "max-encoded-len"], pa
 primitive-types = { default-features = false, version = "0.8.0" }
 serde = { features = ["derive"], optional = true, version = "1.0.101" }
 
+# HydraDX dependencies
+hydra-dx-traits = { git = "https://github.com/galacticcouncil/hydradx-traits", default-features = false }
+
 # Local dependencies
 primitives = { path = "../../primitives", default-features = false }
```
enthusiastmartin, line 26: primitives - I believe this could be now as dev-dependency only as it is used only it tests. Or remove the dependency completely and define the needed types directly in the tests.

This can be the case in other pallets too.

enthusiastmartin (commented):
Not sure yet about the repo. I would prefer to have one common repo for all pallest

lumir-mrkva (changes requested):
move this to the https://github.com/galacticcouncil/warehouse

Roznovjak:
Closing this PR because it is out dated

> move this to the https://github.com/galacticcouncil/warehouse

Can't move it there because it needs to be done in the Basilisk repo. There's nothing to split in the Warehouse repo.

Roznovjak:
I have created new PR : #244

PR #153, opened 2021-09-02
fix: duster dependencies

lumir-mrkva:


enthusiastmartin (dismissed):
For now, it is ok.

I am not sure why it is needed because the dependencies  which are moved from dev are needed only for tests. 

They should not be really needed for release build ( imagine you want to use duster in an environment where there is not the basilisk primitives crate. 

I'll will dig deeper.

enthusiastmartin (approved):
Deps are back in dev-deps. So it is better way.

PR #155, opened 2021-09-07
fix(exchange): fix matching algorithm

enthusiastmartin:


pallets/exchange/src/tests.rs
```diff
@@ -1043,6 +1044,7 @@ fn buy_trade_limits_respected_for_matched_intention() {
 }
 
 #[test]
+#[ignore = "Recent fix changes the behaviour of this test - needs to be adjusted"] // Needs to be adjusted so it tests what it is supposed to.
```
jak-pan, line 1047: What about this?

enthusiastmartin: Updated.  The change caused intentions to be matched slightly differently.

jak-pan (approved):

PR #156, opened 2021-09-08
ci: add clippy

Roznovjak:
## Description
Resolve all clippy warnings and add clippy to github workflow

### DIFF
--- a/pallets/lbp/src/benchmarking.rs
+++ b/pallets/lbp/src/benchmarking.rs
@@ -44,7 +44,7 @@ benchmarks! {
 
 	}: _(RawOrigin::Root, caller.clone(), asset_a, asset_b, duration, WeightCurveType::Linear, true, Fee::default(), caller)
 	verify {
-		assert_eq!(PoolData::<T>::contains_key(&pool_id), true);
+		assert!(PoolData::<T>::contains_key(&pool_id));
 	}
 
 	update_pool_data {
@@ -104,12 +104,12 @@ benchmarks! {
 		ensure!(PoolData::<T>::contains_key(&pool_id), "Pool does not exist.");
 
 		let pool_data = LBP::<T>::pool_data(&pool_id);
-		assert_eq!(pool_data.paused, false);
+		assert!(!pool_data.paused);
 
 	}: _(RawOrigin::Signed(caller), pool_id.clone())
 	verify {
 		let pool_data = LBP::<T>::pool_data(pool_id);
-		assert_eq!(pool_data.paused, true);
+		assert!(pool_data.paused);
 	}
 
 	unpause_pool {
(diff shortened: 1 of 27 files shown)


pallets/lbp/src/benchmarking.rs
```diff
@@ -227,7 +227,7 @@ benchmarks! {
 		ensure!(PoolData::<T>::contains_key(&pool_id), "Pool does not exist.");
 
 		let pool_data = LBP::<T>::pool_data(&pool_id);
-		assert_eq!(pool_data.paused, false);
+		assert!(!pool_data.paused);
```
lumir-mrkva, line 230: i don't like double negation 🙃

lumir-mrkva: and i think that `assert_eq` is more appropriate in this case anyway

pallets/transaction-multi-payment/benchmarking/src/mock.rs
```diff
@@ -163,11 +163,9 @@ impl AssetPairAccountIdFor<AssetId, u64> for AssetPairAccountIdTest {
 		let mut a = asset_a as u128;
 		let mut b = asset_b as u128;
 		if a > b {
-			let tmp = a;
-			a = b;
-			b = tmp;
+			std::mem::swap(&mut a, &mut b)
```
lumir-mrkva, line 166: fancy

runtime/src/lib.rs
```diff
@@ -582,7 +584,7 @@ parameter_types! {
 	// Bond for candidacy into governance
 	pub const CandidacyBond: Balance = 10_000 * DOLLARS;
 	// 1 storage item created, key size is 32 bytes, value size is 16+16.
-	pub const VotingBondBase: Balance = 1 * DOLLARS;
+	pub const VotingBondBase: Balance = DOLLARS;
```
lumir-mrkva, line 587: - [ ] we should really rename DOLLARS already, for example to UNIT

lumir-mrkva (changes requested):

pallets/lbp/src/benchmarking.rs
Roznovjak: There is only one negation :) I agree with Clippy (Clippy suggests to use assert macro). Using `assert_eq` is like using `if condition == true` anti-pattern.

pallets/transaction-multi-payment/benchmarking/src/mock.rs
Roznovjak: ...standard. I would say it is fancy to perform swap on your own.

pallets/lbp/src/benchmarking.rs
lumir-mrkva: ok, you're right, but the second `!` is hard to spot, I actually had to read it multiple times what you';re trying to check, before it was much more obvious

lumir-mrkva: i am not looking at it as predicate, but simply as value check - that's why actually asset_eq makes more sense to me

lumir-mrkva (approved):

Issue #70, opened 2021-06-18
LBP model simulator


lumir-mrkva:
a tool where you can simulate LBP based on predicted sale rate
- ideally, the same params would be applied to many (or range) of sale rate curves

could be an extension of [balancer lbp spreadsheet](https://docs.google.com/spreadsheets/d/1t6VsMJF8lh4xuH_rfPNdT5DM3nY4orF9KFOj2HdMmuY/edit#gid=1392289526) or implemented on top of [lbp info](https://github.com/galacticcouncil/lbp-info)

PR #160, opened 2021-09-15
feat(lbp): add simulation test

Roznovjak:
## Description
Implement a test simulating an LBP event

## Related Issue
Fixes: #70

### DIFF
--- a/pallets/lbp/src/lib.rs
+++ b/pallets/lbp/src/lib.rs
@@ -7,7 +7,7 @@
 use codec::{Decode, Encode};
 use frame_support::sp_runtime::{
 	app_crypto::sp_core::crypto::UncheckedFrom,
-	traits::{AtLeast32BitUnsigned, Hash, Zero, Saturating},
+	traits::{AtLeast32BitUnsigned, Hash, Saturating, Zero},
 	DispatchError, RuntimeDebug,
 };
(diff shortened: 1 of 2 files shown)


Roznovjak:
@jak-pan @lumir-mrkva I have implemented a new test simulating an LBP event under the same conditions as the simulation from the google [sheet](https://docs.google.com/spreadsheets/d/1HPJK5ic-gZebk-tSk99vt79IFbT2d72oGXMIjbGwdCU/edit?usp=sharing). I have also created new [sheet](https://docs.google.com/spreadsheets/d/1jRZcw2gKIkHFz6bzr6azxRkQUE1ldajO8_-GgtllLRQ/edit?usp=sharing) and updated the price calculation in the sheet because the old implementation was using weights from previous blocks. Differences between the results of the simulations are significant.

Roznovjak:
I was able to reduce the difference between the test and the sheet to ~0.3%

enthusiastmartin:
@Roznovjak perhaps it would be better to implement this as part of integration tests ? it uses basilisk-runtime so you'd be working with that instead of mocked runtime. 
It is actually does not require changes in your test - only context which you run it in.  

what do you think ?

lumir-mrkva (approved):

Issue #164, opened 2021-09-20
Change number of nodes in rococo-local config


maht0rz:
Add an extra node to the rococo local config, otherwise polkadot-launch won't produce parachain blocks (at least in my own Basilisk API setup).

### Actual status
The rococo-local `config.json` only contains a single basilisk/parachain node, which seems insufficient to run a local network for purposes of Basilisk API development (indexer/processor)

### Expected status
The rococo-local setup should run with two parachain nodes instead.

PR #165, opened 2021-09-20
feat: Added an extra node to the rococo-local config

maht0rz:
Added an extra parachain node to the rococo-local `config.json`, otherwise the subsquid indexer would not connect normally or blocks would not be produced on the parachain.

## Description
Added an extra parachain node to the rococo-local testnet config with additional cli options.

## Related Issue
Implements: #164

## How Has This Been Tested?
You can test that the rococo-local setup still works by running `polkadot-launch config.json`

### DIFF
--- a/rococo-local/config.json
+++ b/rococo-local/config.json
@@ -48,8 +48,8 @@
         },
         {
           "wsPort": 9989,
-          "port": 31200,
-          "flags": ["--alice", "--rpc-cors=all", "--", "--execution=wasm"]
+          "port": 31201,
+          "flags": ["--bob", "--rpc-cors=all", "--", "--execution=wasm"]
         }
       ]
     }
--- a/rococo-local/config.json
+++ b/rococo-local/config.json
@@ -44,7 +44,12 @@
         {
           "wsPort": 9988,
           "port": 31200,
-          "flags": ["--alice","--", "--execution=wasm"]
+          "flags": ["--alice", "--rpc-cors=all", "--", "--execution=wasm"]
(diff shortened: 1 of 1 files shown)


rococo-local/config.json
```diff
@@ -44,7 +44,12 @@
         {
           "wsPort": 9988,
           "port": 31200,
-          "flags": ["--alice","--", "--execution=wasm"]
+          "flags": ["--alice", "--rpc-cors=all", "--", "--execution=wasm"]
+        },
+        {
+          "wsPort": 9989,
+          "port": 31200,
+          "flags": ["--alice", "--rpc-cors=all", "--", "--execution=wasm"]
```
enthusiastmartin, line 52: This could be Bob?  

`--bob` instead of `--alice`.

rococo-local/config.json
```diff
@@ -44,7 +44,12 @@
         {
           "wsPort": 9988,
           "port": 31200,
-          "flags": ["--alice","--", "--execution=wasm"]
+          "flags": ["--alice", "--rpc-cors=all", "--", "--execution=wasm"]
+        },
+        {
+          "wsPort": 9989,
+          "port": 31200,
```
martinfridrich, line 51: Port should be different than previous node.

maht0rz: fixed

rococo-local/config.json
maht0rz: fixed

martinfridrich (approved):

PR #168, opened 2021-09-23
feat: NFT Auctions

green-jay:
Adds NFT auctions with 3 auction types: English, TopUp and Candle

Early tech specification: https://www.notion.so/Auctions-6227f3d237e44baf8980e83eae74dbb7
https://github.com/w3f/Grants-Program/blob/master/applications/subauction.md

This should be merged to master with a call filter.
Weights and benchmarks to be updated at a later point.


pallets/auctions/src/lib.rs
```diff
@@ -0,0 +1,1356 @@
...
+		} else {
+			Ok(T::CandleDefaultClosingRangesCount::get())
+		}
+	}
+}
+
+///
+/// Implementation of EnglishAuction
+///
+impl<T: Config> NftAuction<T::AccountId, T::AuctionId, BalanceOf<T>, Auction<T>, Bid<T>> for EnglishAuction<T> {
```
lumir-mrkva, line 887: i would advise splitting up the auctions implementations into separate files by the type, similarly as you done with the tests

pallets/auctions/src/lib.rs
```diff
@@ -0,0 +1,1356 @@
...
+		}
+
+		///
+		/// Places a bid on a running auction
+		///
+		/// - validates bid
+		/// - calls the bid() implementation on the given Auction type
+		/// - deposits BidPlaced event
+		///
+		#[pallet::weight(<T as Config>::WeightInfo::bid())]
```
lumir-mrkva, line 440: i imagine that weight for bid can be vastly different for every auction type, how do you solve this issue? is the weight the weight for most expensive bid and you're just refunding if you use cheaper auction type? or can you somehow change weight based on parameter before even extrinsic is dispatched?

lumir-mrkva (commented):
weights doesn't appear to be finished

pallets/auctions/src/benchmarking.rs
```diff
@@ -0,0 +1,35 @@
...
+	}: {
+	} verify {
+	}
+
+	destroy {
+	}: {
+	} verify {
+	}
+
+	bid {
```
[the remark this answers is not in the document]
green-jay: there are just some dummy weights, not real ones. i think creating benchmarks before first round of reviews might be unnecessary overhead, if the code changes a lot after the review, you need to rewrite it. after you guys check it and we agree that the design is done right, we will add those

vikiival:
🐍 
https://github.com/kodadot/nft-gallery/issues/2776

pallets/auctions/src/weights.rs
```diff
@@ -0,0 +1,121 @@
...
+	fn close_auction() -> Weight;
+	fn claim() -> Weight;
+}
+
+/// Weights for pallet_auction using the subauction node and recommended hardware.
+pub struct BasiliskWeight<T>(PhantomData<T>);
+
+impl<T: frame_system::Config> WeightInfo for BasiliskWeight<T> {
+	fn create_auction() -> Weight {
+		(74_809_000 as Weight)
```
apopiak, line 58: How did you determine this weight and why is it so different from the one in the basilisk runtime?

vgantchev: weights are dummies, i.e. random nrs. 
pending benchmarks but curious to first hear your thoughts on general interface design

pallets/auctions/src/types/english.rs
```diff
@@ -0,0 +1,118 @@
...
+
+	///
+	/// Places a bid on an EnglishAuction
+	///
+	/// - removes lock on auction.common_data.last_bid
+	/// - sets lock on new bid
+	/// - updates auction.common_data.last_bid and auction.common_data.next_bid_min
+	/// - if necessary, increases auction end time to prevent sniping
+	///
+	fn bid(&mut self, _auction_id: T::AuctionId, bidder: T::AccountId, bid: &Bid<T>) -> DispatchResult {
```
apopiak, line 40: This and a bunch of other functions should probably be `require_transactional`

pallets/auctions/src/lib.rs
```diff
@@ -0,0 +1,832 @@
...
+		pub fn bid(origin: OriginFor<T>, auction_id: T::AuctionId, amount: BalanceOf<T>) -> DispatchResult {
```
apopiak, line 450: This is not transactional (see comment on english auction bid).

pallets/auctions/src/types/english.rs
```diff
@@ -0,0 +1,118 @@
...
+			<T as crate::Config>::Currency::remove_lock(AUCTION_LOCK_ID, &current_bid.0);
+		}
+		<T as crate::Config>::Currency::set_lock(AUCTION_LOCK_ID, &bidder, bid.amount, WithdrawReasons::all());
+
+		self.common_data.last_bid = Some((bidder, bid.amount));
+		// Set next minimal bid
+		Pallet::<T>::set_next_bid_min(&mut self.common_data, bid.amount)?;
+
+		// Avoid auction sniping
+		Pallet::<T>::avoid_auction_sniping(&mut self.common_data)?;
```
apopiak, line 52: Both of these are fallible and write to storage -> not atomic without `#[transactional]` annotation on the extrinsic

pallets/auctions/src/types/topup.rs
```diff
@@ -0,0 +1,143 @@
...
+	///
+	/// Creates a TopUp Auction
+	///
+	fn create(&self, sender: T::AccountId, auction: &Auction<T>) -> DispatchResult {
+		self.validate_data()?;
+		Pallet::<T>::validate_create(&self.common_data)?;
+		Pallet::<T>::handle_create(sender, auction, &self.common_data)?;
+
+		Ok(())
+	}
```
apopiak, line 16: AFAICT all three implementations are the same. You could think about providing a default implementation in the trait.

pallets/auctions/src/types/topup.rs
```diff
@@ -0,0 +1,143 @@
...
+				Err(Error::<T>::NoChangeOfAuctionType.into())
+			}
+		})
+	}
+
+	///
+	/// Places a bid on an TopUpAuction
+	///
+	fn bid(&mut self, auction_id: T::AuctionId, bidder: T::AccountId, bid: &Bid<T>) -> DispatchResult {
+		// Trasnfer funds to the subaccount of the auction
```
apopiak, line 42:
```suggestion
		// Transfer funds to the subaccount of the auction
```

pallets/auctions/src/lib.rs
```diff
@@ -0,0 +1,832 @@
...
+
+	///
+	/// Helper function which extends auction end time if necessary to prevent auction sniping
+	/// 
+	fn avoid_auction_sniping(common_auction_data: &mut CommonAuctionData<T>) -> DispatchResult {
+		let block_number = <frame_system::Pallet<T>>::block_number();
+		let time_left = common_auction_data
+			.end
+			.checked_sub(&block_number)
+			.ok_or(Error::<T>::TimeUnderflow)?;
```
apopiak, line 796: I would predict a test for this `TimeUnderflow` will fail a `assert_noop!` (see remarks about atomicity).

apopiak (changes requested):
IMO this PR needs a bit of work:
+ I'm pretty confident that at least the `bid` extrinsic is not atomic.
+ If I read the candle auction correctly, a `close` could - in the worst case - lead to a 100 reads and writes because of the `HighestBiddersByAuctionClosingRange` map used.

apopiak:
On benchmarking:
If you want to keep the current API surface/design I would write 3 benchmarks for each extrinsic and then pick the maximum at dispatch and refund according to the actual auction at the end (of the extrinsic execution). If you keep something like the current design for the individual auctions my sense is that the candle auction will be quite a bit more expensive because of the extra reads and writes.

pallets/auctions/src/weights.rs
apopiak: Please don't deploy with random values. At the minimum do a conservative educated guess that you expect to be hard to exploit. (e.g. pallet_xcm has had conservative weights for [some of its extrinsics](https://github.com/paritytech/polkadot/blob/27b839e5d4ffe2c13f19584052d3b3ae8910906a/xcm/pallet-xcm/src/lib.rs#L564-L565) for a while)

pallets/auctions/src/types/candle.rs
```diff
@@ -0,0 +1,277 @@
...
+			Ok(range) => {
+				// <HighestBiddersByAuctionClosingRange<T>>::insert(&auction_id, &range, bidder.clone());
+				<HighestBiddersByAuctionClosingRange<T>>::mutate(
+					&auction_id,
+					&range,
+					|highest_bidder| -> DispatchResult {
+						*highest_bidder = Some(bidder.clone());
+
+						Ok(())
+					},
```
apopiak, line 53: Might be worth comparing this solution to one where you store a BTreeMap in one storage value. Will be more expensive on bid, but much cheaper on close, I imagine.

Also: I'm confused that you just set the `highest_bidder`, might be intentional, though (the details of candle auctions are wiped from my mind again :-D)

apopiak:
About optimizing the candle auction extrinsic weight:
You basically can trade-off between random access weight versus iteration weight.
You get a trade-off dimension that consists of:
1. StorageMap (one key-value entry per data point): fast random access (just mutate the correct key), slow iteration on close
2. chunked StorageMap (putting the data into fewer bins than a naive storage map): can choose the trade-off between fast random access and fast iteration, but higher implementation complexity
3. StorageValue (put all values into one storage value): slower random access (the more values there are), but fast iteration speed

In terms of the chunked map and the StorageValue you can choose between a `Vec` or a `BTreeMap`.

pallets/auctions/src/types/english.rs
vgantchev: Good catch thx (that's my first pallet so i'm still learning)

pallets/auctions/src/types/candle.rs
```diff
@@ -0,0 +1,277 @@
...
+		Ok(())
+	}
+}
+
+impl<T: Config> Pallet<T> {
+  fn choose_random_block_from_range(from: u32, to: u32) -> Result<u32, DispatchError> {
+		ensure!(from < to, Error::<T>::InvalidTimeConfiguration);
+		let mut random_number = Self::generate_random_number(0u32);
+
+		let difference = to.checked_sub(from).ok_or(Error::<T>::Overflow)?;
```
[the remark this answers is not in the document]
vgantchev: i tried using only sp_runtime::ArithmeticError for all math errors but i get the following error:

```^ the trait std::convert::From<sp_runtime::ArithmeticError> is not implemented for pallet::Error<T>```

weird because i look at Martin's Omnipool implementation and i did not see any specific impl for ArithmeticError.

File: pallets/auctions/src/types/english.rs [hunk 13]
```diff
@@ -0,0 +1,118 @@
...
+	/// - sets lock on new bid
+	/// - updates auction.common_data.last_bid and auction.common_data.next_bid_min
+	/// - if necessary, increases auction end time to prevent sniping
+	///
+	fn bid(&mut self, _auction_id: T::AuctionId, bidder: T::AccountId, bid: &Bid<T>) -> DispatchResult {
+		// Lock / Unlock funds
+		if let Some(current_bid) = &self.common_data.last_bid {
+			<T as crate::Config>::Currency::remove_lock(AUCTION_LOCK_ID, &current_bid.0);
+		}
+		<T as crate::Config>::Currency::set_lock(AUCTION_LOCK_ID, &bidder, bid.amount, WithdrawReasons::all());
```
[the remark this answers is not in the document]
vgantchev: That's indeed not the case. I was thinking to simply use reserved_amounts + claim instead of locks, wdyt?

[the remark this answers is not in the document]
apopiak: Yeah, that sounds better to me. You want to reserve for each auction individually and locks can overlap.

apopiak (commented):
some more comments

pallets/auctions/src/tests/topup.rs
```diff
@@ -0,0 +1,872 @@
...
+    assert_eq!(final_auction_account_balance, Zero::zero());
+
+    // Auction data is destroyed
+    assert!(matches!(AuctionsModule::auctions(0), None));
+    assert!(matches!(AuctionsModule::auction_owner_by_id(0), None));
+  });
+}
+
+#[test]
+fn claim_topup_auction_with_winner_should_not_work() {
```
apopiak, line 760: I'm confused about this test. AFAICT this behaviour leads to Bob's funds being locked forever?

vgantchev: https://app.clickup.com/t/2np0jxq

runtime/basilisk/src/lib.rs
```diff
@@ -818,6 +819,37 @@ impl pallet_uniques::Config for Runtime {
...
 
+parameter_types! {
+	pub const AuctionsStringLimit: u32 = 128; // limit length of auction name
+	pub const BidAddBlocks: u32 = 10; // Increase end time to avoid sniping
+	pub const BidStepPerc: u32 = 10; // Next bid step in percent
+	pub const MinAuctionDuration: u32 = 10; // Minimum auction duration
+	pub const BidMinAmount: u32 = 1; // Minimum bid amount
+	pub const CandleDefaultDuration: u32 = 99_356; // Default duration of candle auctions
+	pub const CandleDefaultClosingPeriodDuration: u32 = 72_000; // Default duration of the closing period of a candle auctions
+	pub const CandleDefaultClosingRangesCount: u32 = 100; // Default count of closing periods for candle auctions
```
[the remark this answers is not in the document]
vgantchev: https://app.clickup.com/t/2np0qtw

pallets/auctions/src/lib.rs
```diff
@@ -0,0 +1,852 @@
...
+			<T as Config>::WeightInfo::bid_english()
+				.max(<T as Config>::WeightInfo::bid_topup())
+				.max(<T as Config>::WeightInfo::bid_candle())
+		)]
+		#[transactional]
+		pub fn bid(origin: OriginFor<T>, auction_id: T::AuctionId, amount: BalanceOf<T>) -> DispatchResult {
+			let bidder = ensure_signed(origin)?;
+			let bid = Bid {
+				amount,
+				block_number: frame_system::Pallet::<T>::block_number(),
```
lumir-mrkva, line 473: i would consider using relay chain block numbers for more predictable conversion to time

pallets/auctions/src/lib.rs
```diff
@@ -443,6 +452,8 @@ pub mod pallet {
 				}
 			}
 
+			Self::deposit_event(Event::AuctionDestroyed { id: auction_id });
```
kuchta, line 455: This will not compile...

vgantchev: yup fixed

pallets/auctions/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+
+		/// The default count of closing ranges of a Candle auction
+		#[pallet::constant]
+		type CandleDefaultClosingRangesCount: Get<u32>;
+	}
+
+	#[pallet::storage]
+	#[pallet::getter(fn auctions)]
+	/// Stores on-going and future auctions (closed auctions will be destroyed)
+	pub(crate) type Auctions<T: Config> = StorageMap<_, Twox64Concat, T::AuctionId, Auction<T>, OptionQuery>;
```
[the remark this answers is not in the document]
vgantchev: Good one!

pallets/auctions/src/lib.rs
vgantchev: +100

pallets/auctions/src/lib.rs
```diff
@@ -0,0 +1,882 @@
...
+		/// Auction has reached its ending time (auction.common_data.end is in the past)
+		AuctionEndTimeReached,
+		/// Auction end time has not been reached (auction.common_data.end is in the future)
+		AuctionEndTimeNotReached,
+		/// Auction.common_data.closed can only be set via close() extrinsic
+		CannotSetAuctionClosed,
+		/// Bid amount is invalid
+		InvalidBidPrice,
+		/// Auction count has reached its limit
+		NoAvailableAuctionId,
```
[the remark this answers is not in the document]
vgantchev: 🫡

vgantchev:
Closing this PR in favor of https://github.com/galacticcouncil/Basilisk-node/pull/592

The new PR has isolated English auction and resolved all applicable comments in this PR.

TopUp and Candle auctions are left for a future implementation. Here is a summary of the most important TODOs from this PR:

* Use stronger randomness than randomness_collective_flip for determining candle auction winner (https://github.com/galacticcouncil/Basilisk-node/pull/168/files#r861909731)
* Optimize reads when determining winning range for candle auction (https://github.com/galacticcouncil/Basilisk-node/pull/168#pullrequestreview-957898672)
* Optimize cleaning stored winners by range for candle auction (https://github.com/galacticcouncil/Basilisk-node/pull/168#discussion_r861943603)
* Clean up ReservedAmounts map when closing a TopUp auction which has been won (https://app.clickup.com/t/2np0jxq https://github.com/galacticcouncil/Basilisk-node/pull/168#discussion_r881554036)
* Check test error cases on bid (topup & candle)

PR #170, opened 2021-09-24
refactor: improve multi payment (builder review)

enthusiastmartin:
Builder review suggestions to use origin check rather than having a list of authorities which can add or remove accepted currency.

Also consolidated multi-payment config.


lumir-mrkva (approved):
note: there is no migration that would kill whole authorities storage root, this shouldn't be a problem, because we have no authorities set yet anyway

PR #171, opened 2021-09-24
feat: create testing runtime

Roznovjak:


enthusiastmartin (commented):
regarding the weights in runtime subdir - i believe we can have it in the common runtime and used in testing and basilisk runtime as they would be the same.

pallets/lbp/Cargo.toml
```diff
@@ -17,6 +17,7 @@ substrate-wasm-builder = { git = "https://github.com/paritytech/substrate", bran
 [dependencies]
 codec = { default-features = false, features = ["derive"], package = "parity-scale-codec", version = "2.0.0" }
 primitive-types = { default-features = false, version = "0.8.0" }
+common-runtime = {path = '../../runtime/common', default-features = false}
```
lumir-mrkva, line 20: pallets shouldn't be dependent on the runtime, even the common one, lets use primitives for this

runtime/testing-basilisk/src/xcm.rs
```diff
@@ -0,0 +1,238 @@
+use super::*;
```
lumir-mrkva, line 1: can this be in common?

lumir-mrkva (changes requested):

runtime/testing-basilisk/src/xcm.rs
lumir-mrkva: we should try to reuse as much stuff as possible in common and separate it only when necessary

Roznovjak: I am afraid it will not be possible. Almost everything in the crate depends on items from the runtime crate or on various xcm-related crates.

pallets/lbp/Cargo.toml
Roznovjak: Done. I have also moved `constants.rs` from `common-runtime` to `primitives-crate`.

Roznovjak:
> regarding the weights in runtime subdir - i believe we can have it in the common runtime and used in testing and basilisk runtime as they would be the same.

It would require making the `common-runtime` dependent on all pallets because each pallet defines own `WeightInfo` trait.

README.md
```diff
@@ -57,6 +57,22 @@ Observe Basilisk logs
 multitail 99*.log
 ```
 
+### Testing Runtime
+
+There is also an option to run the testing runtime with less restrictive settings to facilitate testing of new features.
+The following command starts a dev chain, and the testing runtime is used as a runtime for our node.
+```bash
+./target/release/hydra-dx --dev --runtime=testing
```
enthusiastmartin, line 65: basilisk?

README.md
```diff
@@ -57,6 +57,22 @@ Observe Basilisk logs
...
+```
+The testing runtime currently supports only two chain specifications: _dev_ and _local_ testnet.
+Both runtimes store blockchain data in the same directories( e.g. the _dev_ directory is shared for both runtimes 
+started with the `--dev` parameter. That's why it is important to purge chain data when switching to different runtime( note: `--runtime` parameter can't be used when purging chain data)
+
+In the case of starting a testnet using the `polkadot-launch` tool, 
+we don't have an option to communicate to its internal commands that we would like to use the testing runtime.
+To overcome this limitation, rename the binary so it starts with the `testing` prefix, e.g. `testing-basilisk`.
+Such a binary always uses the testing runtime, even if the `--runtime testing` option is not specified.
```
enthusiastmartin, line 74: I feel that there must be a different solution to this.  I personally don't like it - I can already see the questions "why it does not work? -> because you forgot to rename the binary".

Which does not make much sense - it should not be dependent on binary name.

I vaguely remember your explanation why it is necessary. But it still quite clear according to this description.

Perhaps - having possibility to compile in only selected runtime would help ? via a specific feature?

Because we currently compile all runtimes regardless - so i would really like an option to select which runtime i want to use. And that might also resolve this issue ? 
By default, it could compile all runtimes, but option to select a specific one would be nice. 
Just an idea.

Roznovjak: The issue is that when `polkadot-launch` starts a relaychain, it calls `basilisk export-genesis-state` and `basilisk export-genesis-wasm` commands and we don't have an option to tell `polkadot-launch` to add `--runtime testing` argument to those commands. As a result, parathreads are registered with incorrect genesis-state and wasm runtime validation function (for the main runtime) but it tries to spawn parachains with testing runtime.

jak-pan: @Roznovjak Can you try to look at polkadot-launch how much effort will be to add passthrough for flags?

Roznovjak:
The testing runtime has the following settings:

Calls are not filtered
LaunchPeriod = MINUTES
VotingPeriod = MINUTES
Period = 10 * MINUTES

jak-pan:
> The testing runtime has the following settings:
> 
> Calls are not filtered LaunchPeriod = MINUTES VotingPeriod = MINUTES Period = 10 * MINUTES

what is period?

Roznovjak:
> > The testing runtime has the following settings:
> > Calls are not filtered LaunchPeriod = MINUTES VotingPeriod = MINUTES Period = 10 * MINUTES
> 
> what is period?

session length

lumir-mrkva (changes requested):
- [ ] bump versions of everything related

PR #175, opened 2021-09-28
fix: update orml-tokens whitelist accounts

enthusiastmartin:
Updated orml-tokens list of non-dustable accounts. 

It actually uses the duster list as it is duster who handles all the details regarding dust and dusting.

Currenly, on Basilisk - there is only treasury account.

### DIFF
--- a/runtime/src/lib.rs
+++ b/runtime/src/lib.rs
@@ -364,7 +364,7 @@ impl orml_tokens::Config for Runtime {
 	type ExistentialDeposits = AssetRegistry;
 	type OnDust = Duster;
 	type MaxLocks = MaxLocks;
-	type DustRemovalWhitelist = Duster;
+	type DustRemovalWhitelist = pallet_duster::DusterWhitelist<Runtime>;
 }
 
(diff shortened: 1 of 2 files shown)


runtime/src/lib.rs
```diff
@@ -364,7 +364,7 @@ impl orml_tokens::Config for Runtime {
 	type ExistentialDeposits = AssetRegistry;
 	type OnDust = Duster;
 	type MaxLocks = MaxLocks;
-	type DustRemovalWhitelist = Nothing;
+	type DustRemovalWhitelist = Duster;
```
lumir-mrkva, line 367: kinda random that `Duster.contains` actually a set on non-dustable accounts, but I guess is acceptable

lumir-mrkva: i would rather move it to some sub trait like `Duster.dustRemovalWhitelist` that would implement the contains if it's doable

enthusiastmartin: good point - I have not looked at it that way.

It is doable. Will do.

lumir-mrkva (approved):

Issue #111, opened 2021-07-20
DispatchResult is preferred as dispatchable call returns


enthusiastmartin:
Currently all dispatchables return DispatchResultWithPostInfo. It would be better to revert back to DispatchResult

PR #203, opened 2021-10-19
Add tests for English auction

vgantchev:
WIP, more info to follow

### DIFF
--- a/pallets/auctions/src/lib.rs
+++ b/pallets/auctions/src/lib.rs
@@ -200,7 +200,7 @@ pub mod pallet {
 		#[pallet::weight(<T as Config>::WeightInfo::delete_auction())]
 		pub fn delete_auction(origin: OriginFor<T>, id: T::AuctionId) -> DispatchResultWithPostInfo {
 			let sender = ensure_signed(origin)?;
-			let auction = <Auctions<T>>::take(id).ok_or(Error::<T>::AuctionNotExist)?;
+			let auction = <Auctions<T>>::get(id).ok_or(Error::<T>::AuctionNotExist)?;
 			Self::validate_update_delete(&sender, &auction.owner, auction.start)?;
 
 			pallet_uniques::Pallet::<T>::thaw(
--- a/pallets/auctions/src/lib.rs
+++ b/pallets/auctions/src/lib.rs
@@ -220,7 +220,7 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
 		fn on_finalize(now: T::BlockNumber) {
-			Self::conclude_auction(now);
+			Self::conclude_auction(now).unwrap_or_default();
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


pallets/auctions/src/lib.rs
```diff
@@ -194,13 +220,26 @@ pub mod pallet {
 	#[pallet::hooks]
 	impl<T: Config> Hooks<T::BlockNumber> for Pallet<T> {
 		fn on_finalize(now: T::BlockNumber) {
-			Self::conclude_auction(now);
+			Self::conclude_auction(now).unwrap_or_default();
```
green-jay, line 223: what are you trying to do? if you want to pass the result of a function to the outer scope use the question mark. unwrap_or_default should be used only when you actually want to pass the default if unwrap fails

green-jay (changes requested):
remove unwrap and merge

pallets/auctions/src/lib.rs
enthusiastmartin: Note: on finalize cannot fail. And you cannot return any error from there.

enthusiastmartin: And you cannot even return weight consumed by on finalized. That's why you must return return the wieght of on finalize ffrom on_initialized. 

Which i cannot see in this pallet ?! 

You need to benchmarks somehow the weight used in on finalized and return it from on_initialized.

enthusiastmartin (changes requested):
Auction trait:  you can remove associated type AccountId from the trait. You provide it as generic too, so it is useless having it as associated type as well as it does nothing.

You need to benchmark on finalize and return weight from on initialized.

pallets/auctions/src/lib.rs
```diff
@@ -183,9 +199,19 @@ pub mod pallet {
 
 		#[pallet::weight(<T as Config>::WeightInfo::delete_auction())]
 		pub fn delete_auction(origin: OriginFor<T>, id: T::AuctionId) -> DispatchResultWithPostInfo {
```
enthusiastmartin, line 201: It is laready mixed use of DispatchResult and DispatchResultWithPostInfo in this pallet.


Following refactor #111 , it is actaully better to use DispatchResult unless you need specifically the post info.

Applies to each dispatchable in this pallet.

```suggestion
		pub fn delete_auction(origin: OriginFor<T>, id: T::AuctionId) -> DispatchResult {
```

pallets/auctions/src/lib.rs
```diff
@@ -194,13 +220,26 @@ pub mod pallet {
...
 		fn on_finalize(now: T::BlockNumber) {
-			Self::conclude_auction(now);
+			Self::conclude_auction(now).unwrap_or_default();
 		}
 	}
 }
 
 impl<T: Config> Pallet<T> {
-	fn conclude_auction(now: T::BlockNumber) {
+	fn validate_update_delete(
```
enthusiastmartin, line 229: The name is confusing to me . ? What it does - validates, updates or deletes?:-)

According to implementation - i guess it is something like "can_delete" ?  If so, it does not need DispatchResult, simple bool type would suffice.

enthusiastmartin (commented):
Few additional findings.

vgantchev:
@enthusiastmartin thx for review. there is another wip branch with the refactor. lets focus this pr on tests so i can rebase and open the refactor pr

vgantchev:
closing in favour of https://github.com/galacticcouncil/Basilisk-node/pull/233

vgantchev:
closing in favour of #233

Issue #189, opened 2021-10-07
ci: check for version changes before merge


lumir-mrkva:
related to #76

PR #209, opened 2021-10-22
ci: check crate upgrades

Roznovjak:
Fixes: #189


Roznovjak:
The current behavior is that all crates and corresponding versions that have been updated are printed in a comment. As @lumir-mrkva suggested, it would be better to print crates that have been modified but not updated. Unfortunately, I didn't find a github action listing all the crates that have been modified.

Roznovjak:
TODO: 
- [ ] revert [this](https://github.com/galacticcouncil/Basilisk-node/pull/209/commits/58007960206fa9cefb6b0dc2af9c1e704777337f) commit before merging

lumir-mrkva (changes requested):
also please check for runtime spec version upgrades, and that the major version is the same as the spec itself. also testing runtime and regular one should have the same version.

lumir-mrkva (approved):
looks good
- [ ] add boolean output from version-check that states whenever or not runtime was upgraded (to be used in future ci jobs)
- [x] revert testing commits

and then LGTM

Roznovjak:
> looks good
> 
> * [ ]  add boolean output from version-check that states whenever or not runtime was upgraded (to be used in future ci jobs)
> * [ ]  revert testing commits
> 
> and then LGTM

@lumir-mrkva Do you mean to use script's return value as a boolean?

lumir-mrkva:
> > looks good
> > 
> > * [ ]   add boolean output from version-check that states whenever or not runtime was upgraded (to be used in future ci jobs)
> > * [ ]   revert testing commits
> > 
> > and then LGTM
> 
> @lumir-mrkva Do you mean to use script's return value as a boolean?

i meant output from the job like this https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idoutputs so other job can pick up this information

PR #216, opened 2021-10-29
fix: exchange

Roznovjak:
This PR fixes two issues in the `Exchange pallet`:
- overflow of `ExchangeAssetsIntentionCount`
- reservation of an amount which is not required for a trade


enthusiastmartin (commented):
I would like to go through this together to understand better the reasoning.

enthusiastmartin (approved):
Nice catch.

PR #228, opened 2021-11-08
fix: update testing runtime

enthusiastmartin:
Updated testing runtime:
- chain spec to match the runtime chain spec
- orml token adapter


lumir-mrkva (approved):
@enthusiastmartin i assume that we cannot have token adapter in common for some reason

Issue #229, opened 2021-11-08
Improve LBP event order


dexterslabor:
### Detailed Description

The creation of a liquidity bootstrapping pool (LBP) emits the following events (in this order):

1. balances.Endowed
2. balances.Transfer
3. currencies.Transferred
4. tokens.Endowed
5. tokens.Transfer
6. **lbp.PoolCreated**
7. lbp.LiquidityAdded

The improvement is to emit `lbp.PoolCreated` *before* any balance and token transfers events.

### Context

LBP.PoolCreated event contains the `pool_id`. When processing balance and token transfers, it is crucial to limit the processing only to transfers of interest. That's why it is essential to know the `pool_id` beforehand. 
### Possible Implementation

The indexer strictly follows the order as the events are emitted. It is not intended to change this order when processing. Therefore it is not a desired approach to build an exception for this case.

A possible implementation is to change the order of events. The following would pose a viable solution:

1. **lbp.PoolCreated**
1. balances.Endowed
2. balances.Transfer
3. currencies.Transferred
4. tokens.Endowed
5. tokens.Transfer
7. lbp.LiquidityAdded

PR #230, opened 2021-11-08
feat: change the order of emitted events on createPool()

dexterslabor:
## Description
Changed the order of emitted events when calling `createPool()` on the LBP pallet.
## Related Issue
#229


lumir-mrkva:
closes #229

PR #237, opened 2021-11-12
refactor: AssetPairAccountId trait

### LINKED ISSUE #129: refactor: remove AssetPairAccountId from pallet config
### Description
Pallets xyk and lbp ( and hdx in future ) have so called AssetPairAccountId as type in their configs.  Eg: 

`type AssetPairAccountId: AssetPairAccountIdFor<AssetId, Self::AccountId>;`

The trait is is always provided by the pallet! and it is always  implemented by the pallet itself too. 

It does not actually make much sense to have any other implementation. 

This is actually related closely to the pallet itself to generate token names how it needs.

The only part needed is the infix string used when generate a shared asset name .eg "100hdx200". This is used to differentiate between shared asset  created by xyk or lbp pallet which creates shared asset between same assets.

One solution could be that this is part of the AssetPair struct - just need to work you how to do with the infix.

Roznovjak:
Fixes #129

### DIFF
--- a/primitives/src/traits.rs
+++ b/primitives/src/traits.rs
@@ -160,6 +160,6 @@ pub trait ShareTokenRegistry<AssetId, AssetName, Balance, Error>: Registry<Asset
 	}
 }
 
-pub trait AssetPairAccountIdFor<AssetId, PoolId> {
-	fn from_assets(asset_a: AssetId, asset_b: AssetId, identifier: &str) -> PoolId;
+pub trait AssetPairAccountIdFor<AssetId, AccountId> {
+	fn from_assets(asset_a: AssetId, asset_b: AssetId, identifier: &str) -> AccountId;
 }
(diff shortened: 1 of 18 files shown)


primitives/src/traits.rs
```diff
@@ -159,3 +159,7 @@ pub trait ShareTokenRegistry<AssetId, AssetName, Balance, Error>: Registry<Asset
 		}
 	}
 }
+
+pub trait AssetPairAccountIdFor<AssetId, PoolId> {
+	fn from_assets(asset_a: AssetId, asset_b: AssetId, identifier: &str) -> PoolId;
+}
```
enthusiastmartin, line 165:
```suggestion

pub trait AssetPairAccountIdFor<AssetId, AccountId> {
	fn from_assets(asset_a: AssetId, asset_b: AssetId, identifier: &str) -> AccountId;
}
```

enthusiastmartin (commented):
just one suggestion

enthusiastmartin (approved):

enthusiastmartin (approved):

PR #243, opened 2021-11-16
feat: add faucet pallet to testing runtime

enthusiastmartin:


runtime/testing-basilisk/src/lib.rs
```diff
@@ -116,7 +116,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("testing-basilisk"),
 	impl_name: create_runtime_str!("testing-basilisk"),
 	authoring_version: 1,
-	spec_version: 22,
+	spec_version: 23,
```
lumir-mrkva, line 119: @enthusiastmartin i think that both runtime versions should be always in sync even if nothing really changed in other runtime

runtime/testing-basilisk/Cargo.toml
```diff
@@ -42,6 +42,7 @@ pallet-nft = { path = "../../pallets/nft", default-features = false }
 pallet-lbp = { path = "../../pallets/lbp", default-features = false }
 
 pallet-relaychain-info = { git = "https://github.com/galacticcouncil/warehouse", rev="6b399649f0379bc091ff69447b459ce30dd82b00", default-features = false }
+pallet-faucet = { git = "https://github.com/galacticcouncil/warehouse", rev="11cfa1ae1b77ab4a0b86fcf617b18d9e2511c2e8", default-features = false }
```
lumir-mrkva, line 45: we cannot use versions just because it is not merged yet?

lumir-mrkva: wouldn't it better to use version and branch override in the root?

lumir-mrkva:
please always link related `warehouse` pull request

lumir-mrkva:
related https://github.com/galacticcouncil/warehouse/pull/4

runtime/testing-basilisk/src/lib.rs
enthusiastmartin: > @enthusiastmartin i think that both runtime versions should be always in sync even if nothing really changed in other runtime

Ok. I updated the runtime version to match the testing runtime version. Although , I am not convinced with the testing runtime done this way ( but that's another discussion )

runtime/testing-basilisk/Cargo.toml
enthusiastmartin: @lumir-mrkva 
Using vesions in this case won't be much helpful. Because you cant really select a version, there will be always only one version available. 

I would go with the precise revision instead of main branch and version.  3 reasons:

1. To avoid potential error when a pallet/crate is updated with breaking changes but it is not yet reflected in basilisk or hydra.
2. Gives you possibility to work on pallets/crate in warehouse repo without the need to immediately update Basilisk or hydra.
3. If you change version of crate in the warehouse repo - but here it is still specified the previous one - the build fails because it wont find it.

lumir-mrkva: with proper semver it should work nicely

lumir-mrkva (approved):

runtime/testing-basilisk/Cargo.toml
enthusiastmartin: > 
> 
> with proper semver and published crates it should work nicely

yes. with published crates -it would work ok.

Issue #245, opened 2021-11-17
Improve XYK events order


dexterslabor:
# Detailed Description

The creation of a XYK pools emits the following events (in this order):

1. balances.Endowed
1. balances.Transfer
1. currencies.Transferred
1. tokens.Endowed
1. tokens.Transfer
1. currencies.Deposited
1. xyk.PoolCreated

The improvement is to emit xyk.PoolCreated before any balance and token transfers events.

# Context

XYK.PoolCreated event contains the pool_id. When processing balance and token transfers, it is crucial to limit the processing only to transfers of interest. That's why it is essential to know the pool_id beforehand.

# Possible Implementation

The indexer strictly follows the order as the events are emitted. It is not intended to change this order when processing. Therefore it is not a desired approach to build an exception for this case.

A possible implementation is to change the order of events. The following would pose a viable solution:

1. xyk.PoolCreated
1. balances.Endowed
1. balances.Transfer
1. currencies.Transferred
1. tokens.Endowed
1. tokens.Transfer
1. currencies.Deposited

enthusiastmartin:
Fixed in #246

PR #246, opened 2021-11-17
feat: change the order of emitted events for xyk pools

dexterslabor:
Issue #245

### DIFF
--- a/pallets/exchange/src/tests.rs
+++ b/pallets/exchange/src/tests.rs
@@ -38,26 +38,13 @@ fn new_test_ext() -> sp_io::TestExternalities {
 	ext
 }
 
-fn last_event() -> TestEvent {
-	system::Pallet::<Test>::events().pop().expect("Event expected").event
-}
-
 fn expect_event<E: Into<TestEvent>>(e: E) {
-	assert_eq!(last_event(), e.into());
-}
-
-fn last_events(n: usize) -> Vec<TestEvent> {
-	system::Pallet::<Test>::events()
-		.into_iter()
-		.rev()
-		.take(n)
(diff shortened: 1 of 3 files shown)


enthusiastmartin (commented):
The big question is : What happens if one of the transfers fails ? 

`PoolCreated `event is already deposited. 

Would the event be discarded due to the transactional nature of the create call ?  Or not ? 

Personally , I can't tell. It would be good to test it first.

But I'd prefer the deposit event at the very end.

PS: changing the order break all the tests which check the events. And that's lot of those now.

Roznovjak:
> The big question is : What happens if one of the transfers fails ?
> 
> `PoolCreated `event is already deposited.
> 
> Would the event be discarded due to the transactional nature of the create call ? Or not ?
> 
> Personally , I can't tell. It would be good to test it first.
> 
> But I'd prefer the deposit event at the very end.
> 
> PS: changing the order break all the tests which check the events. And there is lot of those now.

I tested it and the event wasn't emitted. Then I did a small research and found out that events are part of the storage. You can find more info here: [frame_system](https://github.com/paritytech/substrate/blob/e670c040d42132da982e156e25a385df213354ed/frame/system/src/lib.rs#L596), [docs](https://docs.substrate.io/v3/runtime/events-and-errors/#depositing-an-event) and [deposit_event()](https://docs.substrate.io/rustdocs/latest/src/frame_system/lib.rs.html#1306)

lumir-mrkva:
@Roznovjak please fix tests, ideally use assert that ignores order of events

lumir-mrkva (approved):

Issue #242, opened 2021-11-16
Liquidity mining - double trouble


martinfridrich:
- [ ] Liq. mining spec
  - [x] spec https://www.notion.so/Liquidity-Mining-codename-double-trouble-2ad34089116047df8eeddba20ecdb383
  - [ ] https://github.com/galacticcouncil/analytics-misc/issues/29
- [x] pallet's internal functions implementation
- [x] public api implementation
  - Admin 
    - [x] create new liq. mining
    - [x] destroy liq. mining
    - [x] withdrawn undistributed rewards
    - [x] add xyk pool
    - [x] update xyk pool settings
    - [x] cancel xyk liq mining
    - [x] resume xyk liq mining
    - [x] remove xyk 
  - User
    - [x] deposit
    - [x] claim
    - [x] withdrawn 
- [x] pallet documentation
- [ ] pallet review
  - [x] testing
  - [x] code review
  - [ ] audit
  - [x] impl. review comments and bugfixing
- [x] benchmarking

martinfridrich:
admin api spec: https://www.notion.so/Liq-mining-Double-trouble-admin-API-293d12f910274091a50dd3d155302c7d

PR #250, opened 2021-11-18
feat: liquidity mining

martinfridrich:
## Description
Implementation of liquidity mining pallet

## Related Issue
#242


pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,754 @@
...
+	/// Return period number from block number now and number of blocks in one period
+	pub fn get_period_number(
+		now: T::BlockNumber,
+		blocks_per_period: T::BlockNumber,
+	) -> Result<T::BlockNumber, Error<T>> {
+		if blocks_per_period.is_one() {
+			return Ok(now);
+		}
+
+		now.checked_div(&blocks_per_period).ok_or(Error::<T>::Overflow)
```
enthusiastmartin, line 515: Is is really necessary to handle 1 as special case ?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,754 @@
...
+			+ Copy
+			+ MaybeSerializeDeserialize
+			+ MaxEncodedLen
+			+ CheckedAdd;
+
+		/// Asset type
+		type CurrencyId: Parameter + Member + Copy + MaybeSerializeDeserialize + Ord + From<u32>;
+
+		/// Currency for transfers
+		type MultiCurrency: MultiCurrencyExtended<
```
enthusiastmartin, line 188: MultiCurrency should be usually sufficient .

Extended only provides `update_balance` which i believe it is not needed

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,754 @@
...
+
+		/// The block number provider
+		type BlockNumberProvider: BlockNumberProvider<BlockNumber = Self::BlockNumber>;
+
+		/// Weight information for extrinsics in this module.
+		type WeightInfo: WeightInfo;
+	}
+
+	#[pallet::error]
+	#[derive(PartialEq)]
```
enthusiastmartin, line 216: This might be redundant.

enthusiastmartin (commented):
Just initial quick review.

I have a feeling that there is a mixture of concrete types and generic types which however expect traits such From<u128> and Into<u32> to be implement by that type ( eg. balance and currency).

If it is defined like that - there are not many options for types anyways and I would consider to set balance and currency types to u128, or u32 respectively. 

This approach will make certain things easier to do - but it limits the "re-use" of the pallet ( but i dont think that's an issue and we should not be worries about that).

eg. XYK is currently done in similar way where balance and currency id is se to u128 and u32.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+				total_rewards,
+				planned_yielding_periods,
+				blocks_per_period,
+				yield_per_period,
+			)?;
+
+			ensure!(
+				T::MultiCurrency::free_balance(reward_currency, &owner) >= total_rewards,
+				Error::<T>::InsufficientRewardCurrencyBalance
+			);
```
enthusiastmartin, line 321: I'd do this before validating the farm data.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+				incentivized_token,
+				max_reward_per_period,
+			);
+
+			<GlobalPoolData<T>>::insert(&pool.id, &pool);
+
+			let pool_account = Self::pool_account_id(pool.id)?;
+			T::MultiCurrency::transfer(reward_currency, &pool.owner, &pool_account, total_rewards)?;
+
+			Self::deposit_event(Event::FarmCreated(pool.id, pool));
```
enthusiastmartin, line 346: Perhaps additional data would be useful. Consider that if FE processor needs to listen some of the events - what date would it be interested in ?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+			updated_at,
+			loyalty_curve,
+			multiplier,
+		}
+	}
+}
+
+#[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebug, TypeInfo)]
+pub struct LoyaltyCurve {
+	b: FixedU128,
```
enthusiastmartin, line 136: Something more descriptive would not be possible instead of just 'b'?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+	) -> Result<Balance, Error<T>> {
+		Ok(yield_per_period
+			.checked_mul_int(total_global_pool_shares)
+			.ok_or(Error::<T>::Overflow)?
+			.min(max_reward_per_period))
+	}
+
+	pub fn update_global_pool(
+		pool_id: PoolId,
+		pool: &mut GlobalPool<T::BlockNumber, Balance, T::CurrencyId, T::AccountId, T::BlockNumber>,
```
enthusiastmartin, line 533: Is it not pool_id part of pool too ?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+	}
+
+	pub fn update_pool(
+		pool_id: PoolId,
+		pool: &mut LiquidityPool<T::BlockNumber, Balance>,
+		rewards: Balance,
+		period_now: T::BlockNumber,
+		global_pool_id: PoolId,
+		reward_currency: T::CurrencyId,
+	) -> Result<(), DispatchError> {
```
enthusiastmartin, line 644: Use DispatchResult as you do in other methods. So it is consistent.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+			Error::<T>::InsufficientBalancdInGlobalPool
+		);
+
+		let global_pool_account = Self::pool_account_id(global_pool_id)?;
+		let pool_account = Self::pool_account_id(pool_id)?;
+		T::MultiCurrency::transfer(reward_currency, &global_pool_account, &pool_account, rewards)
+	}
+
+	pub fn validate_pool_id(pool_id: PoolId) -> Result<(), Error<T>> {
+		if pool_id.le(&Zero::zero()) {
```
enthusiastmartin, line 670: Nitpick: PoolId is u32, it wont be never be < 0. So you can just do 

```suggestion
		if pool_id.is_zero() {
```

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+			total_rewads >= T::MinTotalFarmRewards::get(),
+			Error::<T>::InvalidTotalRewards
+		);
+
+		ensure!(
+			planned_yielding_periods >= T::MinPlannedYieldingPeriods::get(),
+			Error::<T>::InvalidPlannedYieldingPeriods
+		);
+
+		ensure!(!Zero::is_zero(&blocks_per_period), Error::<T>::InvalidBlocksPerPeriod);
```
enthusiastmartin, line 693: Looks more readable.

```suggestion
		ensure!(!blocks_per_period.is_zero(), Error::<T>::InvalidBlocksPerPeriod);
```

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,709 @@
...
+
+		ensure!(!yield_per_period.is_zero(), Error::<T>::InvalidYieldPerPeriod);
+
+		Ok(())
+	}
+
+	pub fn validate_loyalty_curve(c: &Option<LoyaltyCurve>) -> DispatchResult {
```
enthusiastmartin, line 700: This helper function is not very helpful as it is used only once and the implementation is quite straightforward.

i would just inline it.

enthusiastmartin (commented):
Few more comments.

pallets/liquidity-mining/src/lib.rs
martinfridrich: i like this better, but I'll do it :)

pallets/liquidity-mining/src/lib.rs
martinfridrich: I thought that `free_balance()` is doing storage read, that's why I used this order.

pallets/liquidity-mining/src/lib.rs
martinfridrich: I saw it only as a `b`. Maybe @poliwop can suggest better name

poliwop: Ah yes, b is standard notation for "y-intercept" in the U.S.

In this case it's the initial reward, so if b = 0.20, the liquidity mining rewards start at 20% of max and then grow from there the longer the liquidity stays in the pool.

So something like "initial_reward_percentage" would be appropriate.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,829 @@
...
+	}
+
+	#[pallet::error]
+	#[cfg_attr(test, derive(PartialEq))]
+	pub enum Error<T> {
+		/// Math computation overflow
+		Overflow,
+
+		/// Insufficient balance in global pool to transfer rewards to pool
+		InsufficientBalancdInGlobalPool,
```
poliwop, line 209: Typo in variable name

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,829 @@
...
+	fn validate_pool_id(pool_id: PoolId) -> Result<(), Error<T>> {
+		if pool_id.is_zero() {
+			return Err(Error::<T>::InvalidPoolId);
+		}
+
+		Ok(())
+	}
+
+	fn validate_create_farm_data(
+		total_rewads: Balance,
```
poliwop, line 808: Typo in variable name

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1325 @@
...
+					Self::update_global_pool(g_pool, now_period, reward_per_period)?;
+				}
+				g_pool.liq_pools_count = g_pool.liq_pools_count.checked_add(1).ok_or(Error::<T>::Overflow)?;
+
+				let liq_pool_id = Self::get_next_id()?;
+				let pallet_account = Self::account_id();
+				let (nft_class, _) = pallet_nft::Pallet::<T>::do_create_class(
+					pallet_account,
+					ClassType::LiquidityMining,
+					vec![].try_into().unwrap(),
```
enthusiastmartin, line 546: we should not use unwrap. albeit unlikely, it can theoretically cause an issue and audit would again point it out.

Not sure now what do_create_class needs - so i cant suggest anything atm. I will look later.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1325 @@
...
+
+		#[pallet::weight(1000)]
+		#[transactional]
+		pub fn update_liquidity_pool(
+			origin: OriginFor<T>,
+			farm_id: PoolId,
+			asset_pair: AssetPair,
+			multiplier: PoolMultiplier,
+		) -> DispatchResult {
+			//TODO: this fn is WIP, it's not tested
```
enthusiastmartin, line 568: TODO?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1325 @@
...
+						.ok_or(Error::<T>::Overflow)?;
+
+					g_pool
+						.total_shares_z
+						.checked_sub(liq_pool.stake_in_global_pool)
+						.ok_or(Error::<T>::Overflow)?;
+					g_pool
+						.total_shares_z
+						.checked_add(new_stake_in_global_pool)
+						.ok_or(Error::<T>::Overflow)?;
```
enthusiastmartin, line 610: i am bit confused of these calculations. What purpose they have since the result is not used ?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1325 @@
...
+
+			<LiquidityPoolData<T>>::try_mutate_exists(
+				farm_id,
+				amm_account,
+				|maybe_liq_pool| -> DispatchResultWithPostInfo {
+					let liq_pool = maybe_liq_pool.as_mut().ok_or(Error::<T>::LiquidityPoolNotFound)?;
+
+					ensure!(liq_pool.canceled, Error::<T>::LiquidityMiningIsNotCanceled);
+
+					<GlobalPoolData<T>>::try_mutate(farm_id, |maybe_g_pool| -> DispatchResultWithPostInfo {
```
enthusiastmartin, line 674: You can use DispatchResult instead of DispatchWithPostInfo, no ? Like you do everywhere else ?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1325 @@
...
+						T::MultiCurrency::transfer(
+							g_pool.reward_currency,
+							&liq_pool_account,
+							&g_pool_account,
+							unpaid_rew,
+						)?;
+
+						<NftClassData<T>>::try_mutate_exists(
+							liq_pool.nft_class,
+							|maybe_nft_class| -> DispatchResultWithPostInfo {
```
enthusiastmartin, line 695: Samehere, Dispath with posit info not really needed.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1325 @@
...
+			})
+		}
+
+		#[pallet::weight(1000)]
+		#[transactional]
+		pub fn remove_liquidity_pool(
+			origin: OriginFor<T>,
+			farm_id: PoolId,
+			asset_pair: AssetPair,
+		) -> DispatchResultWithPostInfo {
```
enthusiastmartin, line 661: DispatchResult instead DispatchWithPostInfo?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1325 @@
...
+					let deposit = maybe_nft.as_mut().ok_or(Error::<T>::NftDoesNotExist)?;
+
+					let amm_account = T::AMM::get_pair_id(asset_pair);
+					let mut can_destroy_nft_class = true;
+					<LiquidityPoolData<T>>::try_mutate(
+						farm_id,
+						amm_account,
+						|maybe_liq_pool| -> Result<(), DispatchError> {
+							if maybe_liq_pool.is_some() {
+								//This is intetional. This fn should not fail if liq. pool does not
```
enthusiastmartin, line 880:
```suggestion
								//This is intentional. This fn should not fail if liq. pool does not
```

enthusiastmartin (commented):
First few comments. 

I want to go through the math in detail later on.

Other than comments:
- Weight information - it is all set to 1000 atm.
- Few notes and Todos still in code - would be to resolve/answer.

General observation - there seems to be lot of different code paths. It needs to be thoroughly tested. 
Perhaps integration tests would be good to do as well ( it interacts with different parts quite often)

pallets/liquidity-mining/src/lib.rs
martinfridrich: `pallet_nft::do_destory_class()` returns `DispatchWithPostInfo`

pallets/liquidity-mining/src/lib.rs
martinfridrich: This was TODO. It's finished now and ready for review

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1323 @@
...
+				*maybe_g_pool = None;
+
+				Self::deposit_event(Event::FarmDestroyed(farm_id, who));
+				Ok(())
+			})
+		}
+
+		#[pallet::weight(1000)]
+		pub fn withdraw_undistributed_rewards(_origin: OriginFor<T>, _farm_id: PoolId) -> DispatchResult {
+			todo!()
```
enthusiastmartin, line 493: Also this todo. Note that if you call this functions , the todo macro calls panic.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1323 @@
...
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {}
+
+	#[pallet::config]
+	pub trait Config:
+		frame_system::Config + TypeInfo + pallet_nft::Config<ClassType = primitives::nft::ClassType>
+	{
+		type Event: From<Event<Self>> + IsType<<Self as frame_system::Config>::Event>;
+
+		/// Asset type
+		type CurrencyId: Parameter + Member + Copy + MaybeSerializeDeserialize + Ord + From<u32>;
```
enthusiastmartin, line 224: From<u32> is only needed so you can run benchmarking, correct ?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1323 @@
...
+		//b.is_one() is special case - this case is prevented by loyalty curve params validation
+		if FixedPointNumber::is_one(&curve.initial_reward_percentage) {
+			return Ok(FixedU128::one());
+		}
+
+		let denom = curve
+			.initial_reward_percentage
+			.checked_add(&1.into())
+			.ok_or(Error::<T>::Overflow)?
+			.checked_mul(&FixedU128::from(Into::<u128>::into(curve.scale_coef)))
```
enthusiastmartin, line 1045: nitpick.this should work too.

```suggestion
			.checked_mul(&FixedU128::from(curve.scale_coef as u128))
```

pallets/liquidity-mining/src/lib.rs
martinfridrich: it's not needed for now, removed

pallets/exchange/benchmarking/src/mock.rs
```diff
@@ -29,11 +29,11 @@ use sp_runtime::{
 };
 
 use frame_system::EnsureSigned;
+use hydradx_traits::AssetPairAccountIdFor;
```
jak-pan, line 32: probably not for this PR but @enthusiastmartin do we want to move this to gc/warehouse?

pallets/lbp/src/lib.rs
```diff
@@ -768,8 +764,8 @@ impl<T: Config> Pallet<T> {
 				&& pool_data.initial_weight < MAX_WEIGHT
 				&& !pool_data.final_weight.is_zero()
 				&& pool_data.final_weight < MAX_WEIGHT,
-				// TODO people could leak value out the pool if initial weight is < final weight due to fee structure
-				// && pool_data.initial_weight > pool_data.final_weight,
+			// TODO people could leak value out the pool if initial weight is < final weight due to fee structure
```
jak-pan, line 767: Is all of this rustfmt ? If yes can we do separate PR or will it merge conflict?

pallets/liquidity-mining/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,500 @@
...
+	verify {
+		assert!(LiquidityMining::<T>::global_pool(1).is_none());
+	}
+
+	withdraw_undistributed_rewards {
+		let caller = funded_account::<T>("caller", 0);
+
+		init_farm::<T>(1_000_000, caller.clone(), Permill::from_percent(20))?;
+
+		let g_pool_acc = LiquidityMining::<T>::pool_account_id(1).unwrap();
```
jak-pan, line 171: Can you pls use full name for the sake of readability from other reviewers pls?

pallets/nft/src/benchmarking.rs
```diff
@@ -26,164 +27,75 @@ fn dollar(d: u32) -> u128 {
 	d.saturating_mul(100_000_000_000_000)
 }
 
+fn do_create_class<T: Config>() -> (T::AccountId, <T::Lookup as StaticLookup>::Source, BoundedVecOfUnq<T>) {
```
jak-pan, line 30: I think it would be good to comment the functions with reasoning why we think it's the worst case scenario in bench files. This is not related to just this case but to all cases. Creating issue for this.

pallets/lbp/src/lib.rs
jak-pan: If we don't have comprehensive rustfmt settings that check this on PRs we should add this (create new issue) to prevent merge conflicts in the future.

pallets/lbp/src/tests.rs
```diff
@@ -1430,7 +1449,14 @@ fn remove_liquidity_from_not_started_pool_should_work() {
 
 		expect_events(vec![
 			frame_system::Event::KilledAccount(KUSD_BSX_POOL_ID).into(),
-			Event::LiquidityRemoved(KUSD_BSX_POOL_ID, KUSD, BSX, pool_balance_a_before, pool_balance_b_before).into(),
+			Event::LiquidityRemoved(
```
jak-pan, line 1452: Should this kind of test be a part of this pallet (maybe leftover from pre-nft)?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1567 @@
...
+	#[pallet::storage]
+	#[pallet::getter(fn global_pool)]
+	type GlobalPoolData<T: Config> = StorageMap<_, Twox64Concat, PoolId, GlobalPool<T>, OptionQuery>;
+
+	#[pallet::storage]
+	#[pallet::getter(fn liquidity_pool)]
+	type LiquidityPoolData<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		PoolId, //global_pool_id
```
jak-pan, line 473: GlobalPoolId?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1567 @@
...
+				Error::<T>::InsufficientRewardCurrencyBalance
+			);
+
+			let planned_periods =
+				TryInto::<u128>::try_into(planned_yielding_periods).map_err(|_e| Error::<T>::Overflow)?;
+			let max_reward_per_period = total_rewards.checked_div(planned_periods).ok_or(Error::<T>::Overflow)?;
+			let now_period = Self::get_now_period(blocks_per_period)?;
+			let pool_id = Self::get_next_id()?;
+
+			let g_pool = GlobalPool::new(
```
jak-pan, line 536: pls use full variable names for easier readability

jak-pan: *nitpick

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1567 @@
...
+			farm_id: PoolId,
+			asset_pair: AssetPair,
+			multiplier: PoolMultiplier,
+			loyalty_curve: Option<LoyaltyCurve>,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			ensure!(!multiplier.is_zero(), Error::<T>::InvalidMultiplier);
+
+			if let Some(ref c) = loyalty_curve {
```
jak-pan, line 632: curve?

pallets/nft/src/benchmarking.rs
```diff
@@ -26,164 +27,75 @@ fn dollar(d: u32) -> u128 {
...
-		let class_metadata = big_vec;
-		let class_data = ClassData { is_pool:true };
-		let class_id = orml_nft::Pallet::<T>::next_class_id();
-		let price: BalanceOf<T> = u32::MAX.into();
-	}: _(RawOrigin::Signed(caller.clone()), class_metadata, class_data, price)
+		let metadata: BoundedVec<_, _> = vec![0; <T as UNQ::Config>::StringLimit::get() as usize].try_into().unwrap();
+	}: _(RawOrigin::Signed(caller.clone()), Default::default(), metadata)
 	verify {
-		assert_eq!(orml_nft::Pallet::<T>::classes(class_id).iter().count(), 1);
+		assert_eq!(UNQ::Pallet::<T>::class_owner(&T::NftClassId::from(0u32).into()), Some(caller));
```
jak-pan, line 57: Is this the worst case? Should we simulate full class array?

pallets/lbp/src/lib.rs
martinfridrich: Yes, I used `make format` cmd. I'll be merging master into this branch so I can undo this all the format changes if we want

pallets/exchange/benchmarking/src/mock.rs
enthusiastmartin: hydradx traits are already from warehouse.

enthusiastmartin: Yes. I always wondered what it even means:))) but It has been there since the very beginning. 

I will create issue in warehouse

pallets/nft/src/benchmarking.rs
martinfridrich: this should be in changed in nft PR @green-jay

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1644 @@
...
+						ensure!(global_pool.owner == who, Error::<T>::Forbidden);
+
+						global_pool.liq_pools_count =
+							global_pool.liq_pools_count.checked_sub(1).ok_or(Error::<T>::Overflow)?;
+
+						//transfer unpaid rewards back to global_pool
+						let global_pool_account = Self::pool_account_id(global_pool.id)?;
+						let liq_pool_account = Self::pool_account_id(liq_pool.id)?;
+
+						let unpaid_rew =
```
green-jay, line 903:
```suggestion
						let unpaid_reward =
```

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1644 @@
...
+		Self::validate_pool_id(liq_pool_id)?;
+
+		NftInstanceSequencer::<T>::try_mutate(|current_id| {
+			*current_id = current_id.checked_add(1).ok_or(Error::<T>::Overflow)?;
+
+			ensure!(MAX_NFT_INSTANCE_SEQUENCER.ge(current_id), Error::<T>::NftIdOwerflow);
+
+			let mut id_bytes: [u8; 16] = [0; 16];
+
+			id_bytes[..4].copy_from_slice(&liq_pool_id.to_le_bytes());
```
green-jay, line 1262: these magic numbers might be better to put in some const

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1644 @@
...
+			Ok(u128::from_le_bytes(id_bytes))
+		})
+	}
+
+	/// This function decode liq. pool id from nft id
+	fn get_pool_id_from_nft_id(nft_id: NftInstanceIdOf<T>) -> Result<PoolId, Error<T>> {
+		//largets invalid nft id
+		ensure!(4_294_967_296_u128.lt(&nft_id), Error::<T>::InvalidNftId);
+
+		let mut pool_id_bytes = [0; 4];
```
green-jay, line 1274: same here

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1644 @@
...
+		blocks_per_period: BlockNumberFor<T>,
+	) -> Result<PeriodOf<T>, Error<T>> {
+		block.checked_div(&blocks_per_period).ok_or(Error::<T>::Overflow)
+	}
+
+	/// Loyalty multiplier  
+	///
+	// theta = periods/[(initial_reward_percentage + 1) * scale_coef];
+	//
+	// loyalty_multiplier = [theta + (theta * initial_reward_percentage) + initial_reward_percentage]/[theta + (theta * initial_reward_percentage) + 1]
```
green-jay, line 1312: too long line

green-jay (commented):
partly done, still need to finish calc check + benchs + tests

martinfridrich:
> Benchmarks LGTM. Looking at the 7k lines of tests I think it may be better to merge all the happy/unhappy flows from one area with the same setup under one unit test. That might remove a lot of boilerplate code, maybe even reduce it by half. Also one way would be split it to different test files by areas. Bottom line is we need some structure because this is not really easily digestable.

I agree test file is too big but I don't think merging happy/unhappy paths will reduce code that much. I think it will only make test longer and harder to read. "Tests setup" is already extracted to functions. Most of the lines are test data and asserts.  
I'll try split it to multiple files, e.g extract test data to separate file.

lumir-mrkva (changes requested):
- [x] rustfmt
- [ ] sacrifice to clipy
- [ ] bump versions

integration-tests/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "runtime-integration-tests"
-version = "0.2.7"
+version = "0.3.0"
```
enthusiastmartin, line 3: why need of changes in Cargo toml for integrations tests ? it seems that the pallet is not used in the code

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+	use crate::weights::WeightInfo;
+	use frame_system::pallet_prelude::{BlockNumberFor, OriginFor};
+
+	#[pallet::pallet]
+	#[pallet::without_storage_info]
+	pub struct Pallet<T>(_);
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_runtime_upgrade() -> frame_support::weights::Weight {
```
enthusiastmartin, line 246: is this called even when chain is starting ?

enthusiastmartin (approved):
only couple of questions. 

it is huge PR. so it is hard to go through it. But it looks good. Note: I did not verify the logic of LM because i am not familiar with that.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Abbr:
+//  rpvs - reward per valued share
+//  rpz - reward per share in global pool
+
+// Notion spec naming map:
```
jak-pan, line 22: Can we upgrade the spec so that we remove this? If there are equations that require it, add it to Spec. @dexterslabor @martinfridrich  (not blocking)

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+//!
+//! ## Overview
+//!
+//! This pallet provide functionality for liquidity mining program with time incentive(loyalty
+//! factor).
+//! Users are rewarded for each period they stay in liq. mining program.
+//!
+//! Reward per one period is derived from the user's loyalty factor which grows with time(periods)
+//! the user is in the liq. mining and amount of LP shares user locked into deposit.
+//! User's loyalty factor is reseted if the user exits and reenters liq. mining pool.
```
jak-pan, line 41:
```suggestion
//! User's loyalty factor is reset if the user exits and reenters liquidity mining pool.
```

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+//! User can claim rewards without penalties(loyalty factor is no reseted), only withdrawing shares
```
jak-pan, line 42:
```suggestion
//! User can claim rewards resetting loyalty factor, only withdrawing shares
```

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+//! is penalized by loyalty factor reset.
+//! User is rewarded from the next period he enters.
```
jak-pan, line 44:
```suggestion
//! User is rewarded from the next period after he enters.
```

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+//!
+//! User's deposit in liq. mining pool is represented by NFT which is minted for the user when he
```
jak-pan, line 46:
```suggestion
//! User deposit in liquidity mining pool is represented by an NFT which is minted for the user when he
```

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+//! enters liq. mining and is burned when he exits. NFT representing deposit is tradable.
+
+// TODO:
+// add to docs blocks are related to relay chain blocks
```
jak-pan, line 50: pls do

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+pub mod migration;
+pub mod weights;
+
+pub use pallet::*;
+
+type PoolId = u32;
+type GlobalPoolId = PoolId;
+type PoolMultiplier = FixedU128;
+
+//This value is result of: u128::from_le_bytes([255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 255, 0, 0, 0, 0])
```
jak-pan, line 70: Shouldn't this be handled by NFT pallet itself? (constants in the NFT pallet?)

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+pub type NftClassIdOf<T> = <T as pallet_nft::Config>::NftClassId;
+pub type NftInstanceIdOf<T> = <T as pallet_nft::Config>::NftInstanceId;
+
+#[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebugNoBound, TypeInfo)]
+pub struct GlobalPool<T: Config> {
+	id: GlobalPoolId,
+	owner: AccountIdOf<T>,
+	pub updated_at: PeriodOf<T>,
+	total_shares_z: Balance,
+	accumulated_rpz: Balance,
```
jak-pan, line 115: either doc this or rename to something representing the variable

jak-pan: both z and rpz are unreadable. (I'm reading spec so I know now)

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+
+type AccountIdOf<T> = <T as frame_system::Config>::AccountId;
+type AssetIdOf<T> = <T as pallet::Config>::CurrencyId;
+type BlockNumberFor<T> = <T as frame_system::Config>::BlockNumber;
+type PeriodOf<T> = <T as frame_system::Config>::BlockNumber;
+pub type NftClassIdOf<T> = <T as pallet_nft::Config>::NftClassId;
+pub type NftInstanceIdOf<T> = <T as pallet_nft::Config>::NftInstanceId;
+
+#[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebugNoBound, TypeInfo)]
+pub struct GlobalPool<T: Config> {
```
jak-pan, line 110: This should get docs

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+	}
+}
+
+#[derive(Clone, Encode, Decode, PartialEq, Eq, RuntimeDebugNoBound, TypeInfo)]
+pub struct LiquidityPoolYieldFarm<T: Config> {
+	id: PoolId,
+	pub updated_at: PeriodOf<T>,
+	total_shares: Balance,
+	total_valued_shares: Balance,
+	accumulated_rpvs: Balance,
```
jak-pan, line 165: Same here - docs / this naming is hard to understand

pallets/liquidity-mining/src/lib.rs
jak-pan: I believe so

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+		LiquidityPoolUpdated {
+			farm_id: GlobalPoolId,
+			liq_pool_farm_id: PoolId,
+			who: AccountIdOf<T>,
+			asset_pair: AssetPair,
+			multiplier: PoolMultiplier,
+		},
+
+		/// Farm's `accumulated_rpz` was updated.
+		FarmAccRPZUpdated {
```
jak-pan, line 485: docs (the one here doesn't help)

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+
+	/// Liquidity pool yield farm details.
+	#[pallet::storage]
+	#[pallet::getter(fn liquidity_pool)]
+	type LiquidityPoolData<T: Config> = StorageDoubleMap<
+		_,
+		Twox64Concat,
+		GlobalPoolId,
+		Twox64Concat,
+		AccountIdOf<T>, //amm_pool_id   //TODO: create alias amm_pool_id
```
jak-pan, line 522: TODO necessary?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+		/// Create new liquidity mining program with proved parameters.
+		///
+		/// `owner` account have to have at least `total_rewards` balance. This fund will be
+		/// transferred from `owner` to farm account.
+		///
+		/// The dispatch origin for this call must be `T::CreateOrigin`.
+		///
+		/// Parameters:
+		/// - `origin`: account allowed to create new liq. mining program.
+		/// - `total_rewards`: total rewards planned to distribute during. This rewards will be
```
jak-pan, line 549:
```suggestion
		/// - `total_rewards`: total rewards planned to distribute. This will be
```

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+				);
+
+				let amm_pool_id = T::AMM::get_pair_id(asset_pair);
+				ensure!(
+					!<LiquidityPoolData<T>>::contains_key(farm_id, &amm_pool_id),
+					Error::<T>::LiquidityPoolAlreadyExists
+				);
+
+				// update  global pool accumulated RPZ
+				let now_period = Self::get_now_period(global_pool.blocks_per_period)?;
```
jak-pan, line 765: current?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+		/// Emits:
+		/// * `RewardClaimed` if claim happen
+		/// * `SharesWithdrawn` event when successful
+		#[pallet::weight(<T as Config>::WeightInfo::withdraw_shares())]
+		#[transactional]
+		pub fn withdraw_shares(origin: OriginFor<T>, nft_id: NftInstanceIdOf<T>) -> DispatchResultWithPostInfo {
+			let who = ensure_signed(origin)?;
+
+			let nft_owner =
+				pallet_nft::Pallet::<T>::owner(T::NftClass::get(), nft_id).ok_or(Error::<T>::CantFindDepositOwner)?;
```
jak-pan, line 1284: When can this happen?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+	}
+
+	/// This function return new unused `NftInstanceIdOf<T>` with encoded `liq_pool_id` into it or
+	/// error.
+	///
+	/// 4 most significant bytes of `NftInstanceIdOf<T>` are reserved for liq. pool id(`u32`).
+	fn get_next_nft_id(liq_pool_id: PoolId) -> Result<NftInstanceIdOf<T>, Error<T>> {
+		Self::validate_pool_id(liq_pool_id)?;
+
+		NftInstanceSequencer::<T>::try_mutate(|current_id| {
```
jak-pan, line 1448: this feels like it should be a part of nft pallet?

jak-pan:
I've reviewed the core functionality so far only naming nitpicks that can also be addressed later (I'd rather do now since it's probably not hard). I'll continue with logic / tests and benchmarks tmrw

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+			*current_id = current_id.checked_add(1).ok_or(Error::<T>::Overflow)?;
+
+			ensure!(MAX_NFT_INSTANCE_SEQUENCER.ge(current_id), Error::<T>::NftIdOverflow);
```
jak-pan, line 1451: Didn't find tests for this. Do we have them?

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+			id_bytes[POOL_ID_BYTES..].copy_from_slice(&current_id.to_le_bytes()[..NFT_SEQUENCE_BYTES]);
+
+			Ok(u128::from_le_bytes(id_bytes))
+		})
+	}
+
+	/// This function return decoded liq. pool id from `NftInstanceIdOf<T>`
+	fn get_pool_id_from_nft_id(nft_id: NftInstanceIdOf<T>) -> Result<PoolId, Error<T>> {
+		//4_294_967_296(largest invalid nft id) = encoded NftInstanceId from (1,1) - 1
+		ensure!(4_294_967_296_u128.lt(&nft_id), Error::<T>::InvalidNftId);
```
jak-pan, line 1465: I'd also test this

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,1861 @@
...
+
+	/// Global pool details.
+	#[pallet::storage]
+	#[pallet::getter(fn global_pool)]
+	type GlobalPoolData<T: Config> = StorageMap<_, Twox64Concat, GlobalPoolId, GlobalPool<T>, OptionQuery>;
+
+	/// Liquidity pool yield farm details.
+	#[pallet::storage]
+	#[pallet::getter(fn liquidity_pool)]
+	type LiquidityPoolData<T: Config> = StorageDoubleMap<
```
jak-pan, line 517: I would personally set some constants for max amount of pools (any map really) and then benchmark against that max number since maps/double maps might take longer to read/write when there is a lot of entries. Something like 255 should be sufficient.

jak-pan (changes requested):
I have some nitpicks in naming which don't need to be addressed now but we should do have issue for this open. 

One benchmarking concern with maps.
2 tests I'd like to see.

Other than that, very nice job 👍🏻

pallets/liquidity-mining/src/lib.rs
jak-pan: Ok now I see that not but could use better writeup of how this came to be and why.

green-jay:
It should be checked if nft class id is less than ReserveClassIdUpTo from the NFT pallet

lumir-mrkva (approved):

jak-pan (approved):

jak-pan (approved):

PR #257, opened 2021-11-23
feat: dont allow to create xyk pool if  lbp is running

enthusiastmartin:


pallets/lbp/src/lib.rs
```diff
@@ -996,3 +996,11 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, BalanceOf<T>> for Pallet<T
 		T::MaxOutRatio::get()
 	}
 }
+
+pub struct DisallowLBPRunningPool<T>(sp_std::marker::PhantomData<T>);
+
+impl<T: Config> CanCreatePool<AssetId> for DisallowLBPRunningPool<T> {
+	fn can_create(_asset_a: AssetId, _asset_b: AssetId) -> bool {
+		true
```
lumir-mrkva, line 1004: XYK for pair can be only created when LBP already finished and it doesn't have any liquidity anymore

lumir-mrkva: there will be the same condition for unlocking the replay funds

enthusiastmartin: @Roznovjak is implementing this as I am not very familiar with LBP details.

lumir-mrkva (approved):

PR #263, opened 2021-11-25
ci: Testnet deployment automation

### LINKED ISSUE #147: Bug: cannot rerun build
seems like the worker is not created second time when you rerun build with exactly same commit, this is caused most likely with the clash of hash. this could be solved by naming workers by random hash generated in 1st step and propagated through outputs.

fakirAyoub:
## Description
Automate the deployment of k8s components. On push to master:
1. Run build on AWS
2. Build docker image
3. Publish it to ECR
4. Dispatch to K8s deployment repository

## Motivation and Context
Automating the creation of testnet end-to-end after each release

## How Has This Been Tested?
End to end testing cross repositories
Check K8S deployment

fixes #147


.github/workflows/workflow.yml
```diff
@@ -104,6 +103,19 @@ jobs:
...
+        run: sudo snap install docker
+      - name: Sleep for 30 seconds
+        run: sleep 30s
+      - name: build docker image
+        run: sudo docker build -t basilisk:latest docker
+      - name: get ecr credentials
+        run: aws ecr get-login-password --region eu-west-1 | sudo docker login --username AWS --password-stdin 601305236792.dkr.ecr.eu-west-1.amazonaws.com
+      - name: push image to ecr repository
+        run: sudo docker tag basilisk:latest 601305236792.dkr.ecr.eu-west-1.amazonaws.com/basilisk
+      - run: sudo docker push 601305236792.dkr.ecr.eu-west-1.amazonaws.com/basilisk:latest
```
maht0rz, line 118: don't think you'd have to use SUDO on a CI runner, might be potentionally unsafe.

.github/workflows/workflow.yml
```diff
@@ -104,6 +103,19 @@ jobs:
...
+      - name: Copy Binary for docker image creation
+        run: cp target/release/basilisk docker/
+      - name: install docker
+        run: sudo snap install docker
+      - name: Sleep for 30 seconds
+        run: sleep 30s
+      - name: build docker image
+        run: sudo docker build -t basilisk:latest docker
+      - name: get ecr credentials
+        run: aws ecr get-login-password --region eu-west-1 | sudo docker login --username AWS --password-stdin 601305236792.dkr.ecr.eu-west-1.amazonaws.com
```
maht0rz, line 115: can we parametrise all these URLs?

.github/workflows/workflow.yml
```diff
@@ -197,3 +209,59 @@ jobs:
...
+    steps:
+      - uses: actions/setup-ruby@v1
+      - name: Send Webhook Notification
+        if: always()
+        env:
+          JOB_STATUS: ${{ job.status }}
+          WEBHOOK_URL: ${{ secrets.DISCORD_WEBHOOK_URL }}
+          HOOK_OS_NAME: ${{ runner.os }}
+          WORKFLOW_NAME: ${{ github.workflow }}
+        run: |
```
maht0rz, line 252: can we get a notification on successful deployment as well, not only build?

.github/workflows/workflow.yml
fakirAyoub: Agree but it's a temporary machine so

.github/workflows/workflow.yml
fakirAyoub: Yes!

.github/workflows/workflow.yml
fakirAyoub: Yup. Adding that. Will resolve conversation once done

fakirAyoub: Added in the testnet-deployment repo

.github/workflows/workflow.yml
```diff
@@ -197,3 +196,62 @@ jobs:
...
+        run: sleep 60
+      - name: install docker
+        run: sudo snap install docker
+      - name: Sleep for 30 seconds
+        run: sleep 30s
+      - name: build docker image
+        run: sudo docker build -t basilisk:latest docker
+      - name: get ecr credentials
+        run: aws ecr get-login-password --region eu-west-1 | sudo docker login --username AWS --password-stdin ${{ env.ECR_URL }}
+      - name: push image to ecr repository
```
lumir-mrkva, line 226: push it to the dockerhub & github instead, i recommend using action to simplify it like here https://github.com/galacticcouncil/HydraDX-node/blob/master/.github/workflows/docker.yml#L77

lumir-mrkva (changes requested):

.github/workflows/workflow.yml
fakirAyoub: Thanks @lumir-mrkva , will make the change

.github/workflows/workflow.yml
```diff
@@ -197,3 +224,62 @@ jobs:
...
+        run: sudo docker tag basilisk:latest ${{ env.ECR_URL }}/basilisk
+      - run: sudo docker push ${{ env.ECR_URL }}/basilisk:latest
```
lumir-mrkva, line 256: tag the images based on the branch, `master -> latest`, `pr-branch -> parsed-branchname/commit` hash

.github/workflows/workflow.yml
```diff
@@ -197,3 +224,62 @@ jobs:
...
+          JOB_STATUS: ${{ job.status }}
+          WEBHOOK_URL: ${{ secrets.DISCORD_WEBHOOK_URL }}
+          HOOK_OS_NAME: ${{ runner.os }}
+          WORKFLOW_NAME: ${{ github.workflow }}
+        run: |
+          git clone https://github.com/DiscordHooks/github-actions-discord-webhook.git webhook
+          bash webhook/send.sh $JOB_STATUS $WEBHOOK_URL
+        shell: bash
+
+  run-external-deployment:
```
lumir-mrkva, line 275: run deployment only when pipeline is ran in the master branch

docker/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:21.04
```
lumir-mrkva, line 1: move this file to the / please

.github/workflows/workflow.yml
fakirAyoub: @lumir-mrkva can you advise the syntax you have in mind?

docker/Dockerfile
fakirAyoub: Sure, done

.github/workflows/workflow.yml
lumir-mrkva: major problem is taht you cannot use `/` in tag so you have to replace it

so for example if the build is from branch `feat/awesome-feature` it would be tagged as `pr-feat-awesome-feature`

later we can also trigger build on git tags and in that case it would tag it without parsing like `v11.0.1` for example

fakirAyoub: Okay, so, our main build would be the one tagged as `master`. For the branches, I will replace `/` with - before naming.

lumir-mrkva: master branch should be tagged as `latest`, as that is default tag that docker tries to pull

.github/workflows/workflow.yml
```diff
@@ -110,11 +119,31 @@ jobs:
...
+          name: basilisk-raw
+          path: .
+      - name: S3 Upload basilisk-raw.json
+        uses: tpaschalis/s3-sync-action@master
 
   destroy-runner-instance:
     name: destroy runner
     runs-on: ubuntu-latest
-    needs: build
+    needs: [build, upload-config-to-s3, build-and-publish-docker-image]
```
lumir-mrkva, line 146: you don't have to wait for `upload-config-to-s3` as they're not running on that runner

```suggestion
    needs: build-and-publish-docker-image
```

.github/workflows/workflow.yml
```diff
@@ -100,6 +99,16 @@ jobs:
...
         run: ./target/release/basilisk --version
+      - name: Get basilisk-raw json
+        run: |
+          cp target/release/basilisk target/release/testing-basilisk
+          chmod +x ./target/release/testing-basilisk
+          ./target/release/testing-basilisk build-spec --chain testnet-k8s --raw > basilisk-raw.json
+      - name: Upload config file
+        uses: actions/upload-artifact@v2
+        with:
+          name: basilisk-raw
```
lumir-mrkva, line 110: rename it to `basilisk-testnet-chainspec.json`

.github/workflows/workflow.yml
```diff
@@ -152,120 +181,59 @@ jobs:
...
+      ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+      ECR_URL: 601305236792.dkr.ecr.eu-west-1.amazonaws.com
+      AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
     steps:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
           ref: ${{ github.event.pull_request.head.sha }}
+      - name: sleep 60
```
lumir-mrkva, line 198: why sleep?

.github/workflows/workflow.yml
```diff
@@ -152,120 +181,59 @@ jobs:
...
-      - run: npm install
-        name: Setup wait script
-        working-directory: scripts/wait-for-inclusion
-      - run: yarn start ../../rococo-local/config.json & cd ../wait-for-inclusion && npm start 200
-        name: Launch rococo local & wait for inclusion
-        working-directory: scripts/polkadot-launch
+          path: .
+      - name: Add execution for basilisk
+        run: chmod +x basilisk
+      - name: Sleep for 60 seconds
```
lumir-mrkva, line 213: why sleep?

.github/workflows/workflow.yml
```diff
@@ -152,120 +181,59 @@ jobs:
...
+        run: sleep 60
+      - name: install docker
```
lumir-mrkva, line 215: docker should be already in ami

.github/workflows/workflow.yml
lumir-mrkva: @fakirAyoub this is still an issue

.github/workflows/workflow.yml
lumir-mrkva: @fakirAyoub still not resolved

.github/workflows/workflow.yml
```diff
@@ -152,120 +181,59 @@ jobs:
...
+        run: sleep 30s
+      - name: build docker image
+        run: sudo docker build -t basilisk:latest .
+      - name: get ecr credentials
+        run: aws ecr get-login-password --region eu-west-1 | sudo docker login --username AWS --password-stdin ${{ env.ECR_URL }}
+      - name: push image to ecr repository
+        run: sudo docker tag basilisk:latest ${{ env.ECR_URL }}/basilisk
+      - run: sudo docker push ${{ env.ECR_URL }}/basilisk:latest
 
-  test-upgrade-runtime:
```
lumir-mrkva, line 197: why is this removed?

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:21.04
+
+RUN apt update && apt install -y ca-certificates
+
+workdir /basilisk
```
lumir-mrkva, line 5:
```suggestion
WORKDIR /basilisk
```


note that you'll have to change following command accordingly (remove basilisk from path)

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+    chown -R basilisk:basilisk /basilisk
+    
+USER basilisk
+ADD ./basilisk /basilisk/testing-basilisk
+
+EXPOSE 30333 9933 9944
+
+VOLUME ["/data"]
+
+entrypoint [ "/basilisk/testing-basilisk" ]
```
lumir-mrkva, line 19: ENTRYPOINT, but i prefer CMD instead if possible

Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM ubuntu:21.04
```
lumir-mrkva, line 1: use slimmer image

lumir-mrkva (changes requested):

Dockerfile
lumir-mrkva: note that with entrypoint you cannot use basilisk, you can only execute testing basilisk

Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+
+workdir /basilisk
+
+RUN useradd -m -u 1000 -U -s /bin/sh -d /basilisk basilisk && \
+    mkdir -p /basilisk/.local/share/basilisk && \
+    ln -s /basilisk/.local/share/basilisk /data && \
+    chown -R basilisk:basilisk /basilisk
+    
+USER basilisk
+ADD ./basilisk /basilisk/testing-basilisk
```
lumir-mrkva, line 13: image should be able to run both basilisk and testing basilisk, eg. basilisk should be copied and then testing basilisk hard linked

Dockerfile
lumir-mrkva: disregard, ubuntu is using slim images by default

.github/workflows/workflow.yml
fakirAyoub: We do, because the file is uploaded from this machine, when it start destroying we cannot get the file from it anymore

Dockerfile
martinfridrich: Why do you prefer CMD? You can override both. Entrypoint is easier to be used by para registration script. If you change it to CMD, don't forget to update para. registration script.

.github/workflows/workflow.yml
fakirAyoub: for some reason, without doing this install step, I have an error saying that Docker cannot be found.

Weird error since the build jobs use it correctly... Can we address this in another PR?

.github/workflows/workflow.yml
fakirAyoub: was for test only, adding it before merge

Dockerfile
fakirAyoub: let's keep ENTRYPOINT for now

.github/workflows/workflow.yml
lumir-mrkva: i just thought it is already there as i am using it here https://github.com/galacticcouncil/HydraDX-node/blob/master/.github/workflows/docker.yml#L46

lumir-mrkva: mby those actions before ensure that docker is installed already? or the ami is setup differently?

Dockerfile
lumir-mrkva: ok

.github/workflows/workflow.yml
fakirAyoub: I have to check the AMI for Basilisk

.github/workflows/workflow.yml
lumir-mrkva: thats not true, you are uploading it from github ubuntu, see https://github.com/galacticcouncil/Basilisk-node/pull/263/files/8ba96bc1635496a9240397f0b28524889d67b934#diff-126bf89616b7daa3d14ebc882ad18666aaf1c3dae888c4ba306a66ec80758bc1R126

lumir-mrkva: the file is already uploaded to github artifacts even before build job finishes

Dockerfile
lumir-mrkva: @martinfridrich can the normal basilisk be the default entrypoint then at least ? :)

martinfridrich: @lumir-mrkva sure, we can make special case for basilisk in para reg. srcipt. Basilisk will be better entrypoint if this images will be public. I just wanted to have unified way to call subcmds in docker images by script. This change have to  be reflected into para reg. script

lumir-mrkva (approved):

lumir-mrkva:
@martinfridrich please merge this when rest of scripts are ready and the docker file default entrypoint changed

lumir-mrkva:
fixes #147

PR #272, opened 2021-11-29
fix: restricted vesting calls

jak-pan:


lumir-mrkva:
why is this necessary? we are already allowing to vest tokens to sudo or Treasury account https://github.com/galacticcouncil/Basilisk-node/blob/f9f019d18a0a41cbf4b66a61023591ebe60d464b/runtime/basilisk/src/lib.rs#L712

lumir-mrkva (approved):

lumir-mrkva (changes requested):
- [ ] vesting benchmarks

lumir-mrkva (approved):

PR #274, opened 2021-12-02
ci: runtime upgrade test

lumir-mrkva:


scripts/upgrade-runtime/index.js
```diff
@@ -0,0 +1,101 @@
...
+  await waitForBlock(3);
+
+  console.log('submitting runtime upgrade');
+  watchEventsFromSections(["sudo", "parachainSystem"]);
+  await waitTx(sudo);
+
+  await new Promise(resolve => onEvent('parachainSystem.ValidationFunctionApplied', resolve));
+  api.rpc.chain.subscribeNewHeads(async header => {
+    const newSpec = getSpecVersion();
+    if (specVersion < newSpec) {
```
jak-pan, line 81: I'd try waiting here for a few blocks to check if the chain is running after upgrade

jak-pan: Maybe do some tx?

lumir-mrkva: i thought that chain continuing to produce blocks is sufficient enough

lumir-mrkva: it actually takes few block before api initialises into the new spec

lumir-mrkva: added tx anyway

runtime/basilisk/src/lib.rs
```diff
@@ -106,7 +106,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("basilisk"),
 	impl_name: create_runtime_str!("basilisk"),
 	authoring_version: 1,
-	spec_version: 26,
+	spec_version: 27,
```
jak-pan, line 109: We don't need to bump this since there is no change to runtime. Is there?

lumir-mrkva: not really, this was changed for testing purposes

jak-pan (commented):
LGTM

jak-pan (approved):

PR #283, opened 2021-12-07
refactor: NFT

green-jay:
NFT pallet that provides basic generic interface to manipulate with NFTs
- create class
- mint instance
- transfer instance
- burn instance
- destroy class

It is possible to directly call extrinsics from this pallet but also provides public interface (do\__method_) for pallets that are built on top of it.

tech spec: https://www.notion.so/NFT-Technical-specification-f387682d6ef34b6091f5607bfa912a19

**REVIEW OF PALLET-UNIQUES IS NOT NEEDED**

_Note: NFT pallet has been migrated to wrap FRAME Uniques instead of ORML. Because there is a couple of slight customizations to Uniques, it was necessary to fork the pallet. There is a PR open already, hopefully it will get to substrate codebase soon enough so we can use the Uniques pallet natively.
https://github.com/paritytech/substrate/pull/10529
Update 20211231: PR merged, waiting for release_
_Update 20220118: Can be used natively from v0.9.15_
_Update 20220204: Uniques fork removed and is used natively_


lumir-mrkva:
blocked by polkadot 0.9.15 upgrade

lumir-mrkva (changes requested):
- [x] #321 
- [x] use substrate uniques pallet instead of local copy

pallets/nft/src/lib.rs
```diff
@@ -281,71 +223,145 @@ pub mod pallet {
...
-				Self::deposit_event(Event::TokenMintedLockToggled(
-					account.clone(),
-					token_id.0,
-					token_id.1,
-					token.data.locked,
-				));
-			}
-			Ok(())
-		})
+	pub fn instance_owner(class_id: T::NftClassId, instance_id: T::NftInstanceId) -> Option<T::AccountId> {
```
lumir-mrkva, line 276: instance_owner -> owner

keep it simple

pallets/nft/src/lib.rs
```diff
@@ -281,71 +223,145 @@ pub mod pallet {
...
+			deposit_info.1,
+			pallet_uniques::Event::Created(class_id.into(), owner.clone(), owner),
+		)?;
+
+		Classes::<T>::insert(class_id, ClassInfo { class_type, metadata });
+
+		Ok((class_id, class_type))
+	}
+
+	pub fn do_mint(
```
lumir-mrkva, line 306: i would prefer to allow to specify your own id when minting NFT's, especially when uniques pallet supports this

pallets/nft/src/lib.rs
```diff
@@ -281,71 +223,145 @@ pub mod pallet {
...
-					token.data.locked,
-				));
-			}
-			Ok(())
-		})
+	pub fn instance_owner(class_id: T::NftClassId, instance_id: T::NftInstanceId) -> Option<T::AccountId> {
+		pallet_uniques::Pallet::<T>::owner(class_id.into(), instance_id.into())
+	}
+
+	pub fn do_create_class(
```
lumir-mrkva, line 280: i would prefer to allow specifying your own id when creating class, we can have some range reserved for runtime nfts

martinfridrich:
Please move events to `do_...()` functions.  FE need events about minting/burning NFTs called from liq. mining pallet

runtime/basilisk/src/lib.rs
```diff
@@ -724,6 +720,38 @@ impl orml_vesting::Config for Runtime {
...
+
+parameter_types! {
+	pub const ClassDeposit: Balance = 100 * UNITS; // 100 UNITS deposit to create asset class
+	pub const InstanceDeposit: Balance = 100 * UNITS; // 100 UNITS deposit to create asset instance
+	pub const KeyLimit: u32 = 256;	// Max 256 bytes per key
+	pub const ValueLimit: u32 = 1024;	// Max 1024 bytes per value
+	pub const UniquesMetadataDepositBase: Balance = 100 * UNITS;
+	pub const AttributeDepositBase: Balance = 10 * UNITS;
+	pub const DepositPerByte: Balance = UNITS;
+	pub const UniquesStringLimit: u32 = 128;
```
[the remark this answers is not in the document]
green-jay: set to 60 for now

pallets/nft/src/lib.rs
green-jay: how about initializing the sequencer to start from e.g. 100 000 000?

green-jay:
> Please move events to `do_...()` functions. FE need events about minting/burning NFTs called from liq. mining pallet

done

primitives/src/lib.rs
```diff
@@ -79,6 +80,94 @@ pub struct ExchangeIntention<AccountId, Balance, IntentionID> {
...
+}
+
+pub mod nft {
+	use super::*;
+
+	#[derive(Encode, Decode, Eq, PartialEq, Copy, Clone, RuntimeDebug, TypeInfo)]
+	#[cfg_attr(feature = "std", derive(Serialize, Deserialize))]
+	pub enum ClassType {
+		Bare = 0_isize,
+		Marketplace = 1_isize,
```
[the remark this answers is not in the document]
green-jay: I wanted to have a default fallback category in case marketplace nfts will behave differently (e.g. some permission restrictions) in future. Marketplace can be the default type and can be named also differently. Any suggestions?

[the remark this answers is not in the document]
lumir-mrkva: it will be really hard to migrate between classes, so most likely you will have to enable this functionality for more classes. so marketplace wont be most probably 1:1 but 1:n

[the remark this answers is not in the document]
lumir-mrkva: therefore we don't really need this class to exist in my opinion, it is just misleading

pallets/nft/src/lib.rs
lumir-mrkva: not for instance id

lumir-mrkva: you should be completely in control of your class instance id's

lumir-mrkva: i dont think that you can have both, custom id and sequencer, custom id has to be default, but you can attach sequencer of top of it if you wish

green-jay: Not my wish :) having auto ID was a strong requirement from Koda so I added it. But still not convinced custom ID is a must.

pallets/nft/src/lib.rs
```diff
@@ -2,277 +2,218 @@
...
-		/// Returns the bond amount
+		/// Removes a class from existence
 		///
 		/// Parameters:
-		/// - `class_id`: unique identificator of a class
+		/// - `class_id`: The identifier of the asset class to be destroyed.
 		#[pallet::weight(<T as Config>::WeightInfo::destroy_class())]
 		#[transactional]
-		pub fn destroy_pool(origin: OriginFor<T>, class_id: ClassIdOf<T>) -> DispatchResult {
+		pub fn destroy_class(origin: OriginFor<T>, class_id: T::NftClassId) -> DispatchResultWithPostInfo {
```
lumir-mrkva, line 205: do we need `DispatchResultWithPostInfo`, why not use just a `DispatchResult`?

green-jay: For us it would be but Uniques requires it. They calculate destroy weight based on class content. Therefore we have to comply.

.github/workflows/workflow.yml
```diff
@@ -299,12 +299,19 @@ jobs:
           name: basilisk_runtime.compact.compressed.wasm
           path: scripts/upgrade-runtime
       - uses: dawidd6/action-download-artifact@v2
-        name: Download basilisk binary from master
+        name: Download basilisk binary from master (legacy)
```
jak-pan, line 302: what is the reasoning here?

pallets/nft/src/benchmarking.rs
```diff
@@ -26,164 +28,80 @@ fn dollar(d: u32) -> u128 {
...
-		let class_metadata = big_vec;
-		let class_data = ClassData { is_pool:true };
-		let class_id = orml_nft::Pallet::<T>::next_class_id();
-		let price: BalanceOf<T> = u32::MAX.into();
-	}: _(RawOrigin::Signed(caller.clone()), class_metadata, class_data, price)
+		let metadata: BoundedVec<_, _> = vec![0; <T as UNQ::Config>::StringLimit::get() as usize].try_into().unwrap();
+	}: _(RawOrigin::Signed(caller.clone()), CLASS_ID_0.into(), Default::default(), metadata)
 	verify {
-		assert_eq!(orml_nft::Pallet::<T>::classes(class_id).iter().count(), 1);
+		assert_eq!(UNQ::Pallet::<T>::class_owner(&T::NftClassId::from(CLASS_ID_0).into()), Some(caller));
```
jak-pan, line 63: Is this the worst case? Is adding to full vec more expensive? > https://substrate.stackexchange.com/questions/88/is-there-a-limit-to-the-size-of-a-map-in-pallet-storage

green-jay: It is adding to a map, not to a vec. Testing adding stuff to a nearly full map might be an overkill. There should be another incentives to keep the map size reasonable, in this case there is a configurable bond to create a class that exists until the class is destroyed.

.github/workflows/workflow.yml
green-jay: i think the goal here is to properly test a runtime upgrade from old to new version instead of new->bumped new?

Issue #284, opened 2021-12-07
LBP validate fee collectors


lumir-mrkva:
currently, it's possible to have a fee collector that collects the same asset from the multiple running LBP's, this causes a problem because we rely on a single LBP collector lock to track the amount collected. so in this case multiple LBP's are essentially rewriting the same value.

this should be solved by restricting unique `(fee collector, collected asset)` pair. this also means that single fee collector can still collect fees from multiple lbps, but collected assets have to be distinct.

implementation would be adding a new map in pallet storage where the key is this pair and a boolean value.

```
(fee_collector, collected_asset_id) -> boolean
```

note that a collected asset cannot change after LBP creation. so only fee collector changes have to be checked with always the same collected asset.

when lbp is created
- validate that fee collector is not set yet in the map
- set fee collector in the map

when fee collector in lbp is changed
- validate that the new fee collector is not set yet
- unset old fee collector, set new fee collector

when lbp funds are withdrawn (lbp is removed)
- unset fee collector

this will done on based on https://github.com/galacticcouncil/Basilisk-node/pull/231 branch

PR #287, opened 2021-12-08
feat: Add validation on existing fee collector

vgantchev:
Resolves https://github.com/galacticcouncil/Basilisk-node/issues/284

### DIFF
--- a/pallets/lbp/src/lib.rs
+++ b/pallets/lbp/src/lib.rs
@@ -347,7 +347,6 @@ pub mod pallet {
 	/// Storage used for tracking existing fee collectors
 	/// Not more than one fee collector per asset possible
 	#[pallet::storage]
-	#[pallet::getter(fn fee_collectors)]
 	pub type FeeCollectorWithAsset<T: Config> =
 		StorageDoubleMap<_,
 			Blake2_128Concat,
@@ -422,7 +421,10 @@ pub mod pallet {
 
 			ensure!(!Self::exists(asset_pair), Error::<T>::PoolAlreadyExists);
 
-			Self::validate_fee_collector(fee_collector.clone(), asset_a)?;
+			ensure!(
+				!<FeeCollectorWithAsset<T>>::contains_key(fee_collector.clone(), asset_a),
+				Error::<T>::FeeCollectorWithAssetAlreadyUsed
+			);
 
 			ensure!(
 				T::MultiCurrency::free_balance(asset_a, &pool_owner) >= asset_a_amount,
@@ -772,21 +774,6 @@ impl<T: Config> Pallet<T> {
 		);
 
 		ensure!(!pool_data.fee.denominator.is_zero(), Error::<T>::FeeAmountInvalid);
-
-		Ok(())
-	}
-
-	/// Validates fee collector
-	/// Not more than one fee collector per asset
-	/// This is done to prevent race conditions with locks on the same asset
-	fn validate_fee_collector(
-		fee_collector: T::AccountId,
-		asset: AssetId
-	) -> DispatchResult {		
-		ensure!(
-			!<FeeCollectorWithAsset<T>>::contains_key(fee_collector, asset),
-			Error::<T>::FeeCollectorWithAssetAlreadyUsed
-		);
 
 		Ok(())
 	}
--- a/pallets/lbp/src/lib.rs
+++ b/pallets/lbp/src/lib.rs
@@ -533,10 +533,18 @@ pub mod pallet {
 
 				// Handle update of fee collector - validate and replace old fee collector
 				if let Some(updated_fee_collector) = fee_collector {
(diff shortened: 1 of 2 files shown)


pallets/lbp/src/lib.rs
```diff
@@ -295,6 +295,9 @@ pub mod pallet {
 
 		/// Amount is less than minimum trading limit.
 		InsufficientTradingAmount,
+
+		/// Not more than one fee collector per asset id
+		FeeCollectorAlreadyExists
```
lumir-mrkva, line 300: `FeeCollectorWithAssetAlreadyUsed` could be little bit more accurate

lumir-mrkva (changes requested):

pallets/lbp/src/lib.rs
```diff
@@ -341,6 +344,17 @@ pub mod pallet {
 	pub type PoolData<T: Config> =
 		StorageMap<_, Blake2_128Concat, PoolId<T>, Pool<T::AccountId, T::BlockNumber>, OptionQuery>;
 
+	/// Storage used for tracking existing fee collectors
+	/// Not more than one fee collector per asset possible
+	#[pallet::storage]
+	#[pallet::getter(fn fee_collectors)]
+	pub type FeeCollectors<T: Config> =
```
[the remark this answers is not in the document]
vgantchev: but if i remove the public getter, how can I test that the value is set in store?

pallets/lbp/src/lib.rs
```diff
@@ -341,6 +344,17 @@ pub mod pallet {
...
+		StorageDoubleMap<_,
+			Blake2_128Concat,
+			T::AccountId,
+			Blake2_128Concat,
+			AssetId, bool, OptionQuery>;
```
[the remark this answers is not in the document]
enthusiastmartin: Just a note to be aware : you get the default value of bool if such entry does not exist in the map. In case of bool - it is false, that's probably what you want.

however, as i can see - you dont use the get by value at all anyway.

pallets/lbp/src/lib.rs
```diff
@@ -513,7 +530,17 @@ pub mod pallet {
...
 				pool.fee = fee.unwrap_or(pool.fee);
-				pool.fee_collector = fee_collector.unwrap_or_else(|| pool.fee_collector.clone());
+
+				// Handle update of fee collector - validate and replace old fee collector
+				if let Some(updated_fee_collector) = fee_collector {
+					Self::validate_fee_collector(updated_fee_collector.clone(), pool.assets.0)?;
+
+					<FeeCollectorWithAsset<T>>::remove(pool.fee_collector.clone(), pool.assets.0);
+					<FeeCollectorWithAsset<T>>::insert(updated_fee_collector.clone(), pool.assets.0, true);
+
```
enthusiastmartin, line 540: With the following refactor, you get one less storage read ( the check in validate).

Also note that the clone calls have disappeared. there is no need to do that. 

And this way, you get the benefit of ValueQuery too.

```suggestion
					FeeCollectorWithAsset::<T>::try_mutate(
						&updated_fee_collector,
						pool.assets.0,
						|collector| -> DispatchResult {
							ensure!(!*collector, Error::<T>::FeeCollectorWithAssetAlreadyUsed);

							<FeeCollectorWithAsset<T>>::remove(&pool.fee_collector, pool.assets.0);
							*collector = true;

							Ok(())
						},
					)?;

```

pallets/lbp/src/lib.rs
```diff
@@ -740,6 +768,21 @@ impl<T: Config> Pallet<T> {
 		Ok(())
 	}
 
+	/// Validates fee collector
+	/// Not more than one fee collector per asset
+	/// This is done to prevent race conditions with locks on the same asset
+	fn validate_fee_collector(
+		fee_collector: T::AccountId,
```
enthusiastmartin, line 775: If you still need this helper function after the change to use try_mutate. 

I'd consider either remove it completely because it is just one ensure and inline it.

Or use a reference for the collector param, so no need to copy everywhere.

pallets/lbp/src/lib.rs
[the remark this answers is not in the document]
lumir-mrkva: we can keep it required, but you already query it without it here https://github.com/galacticcouncil/Basilisk-node/pull/287/files#diff-5a515a229f305ec8027975f033a19da4ffcedc912649a188de42b8b7ee0f1f48R1390

pallets/lbp/src/lib.rs
lumir-mrkva: benefit of the function can be testability, where you can check that this behaves correctly with some storage state, without any complex storage setup

lumir-mrkva: where you would just set storage and check if the result is expected, without any pool creation etc ...

vgantchev:
thank you masters 🙏

lumir-mrkva (approved):

PR #290, opened 2021-12-16
feat: use spot price in multi payment instead of pool trade

enthusiastmartin:


pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -121,6 +121,10 @@ pub mod pallet {
 		/// Accepted currency removed
 		/// [currency]
 		CurrencyRemoved(AssetId),
+
+		/// Transaction fee paid in non-native currency
+		/// [Account, Currency, Native fee amount, Non-native fee amount]
+		FeeTransferred(T::AccountId, AssetId, Balance, Balance),
```
lumir-mrkva, line 127: `FeeWithdrawn` is a better name as transfer event should always state destination as well

pallets/transaction-multi-payment/src/lib.rs
```diff
@@ -345,23 +349,21 @@ impl<T: Config> CurrencySwap<<T as frame_system::Config>::AccountId, Balance> fo
...
 			CORE_ASSET_ID => Ok(PaymentSwapResult::Native),
 			currency => {
-				if T::AMMPool::exists(AssetPair {
+				let amount = if T::AMMPool::exists(AssetPair {
 					asset_in: currency,
 					asset_out: CORE_ASSET_ID,
 				}) {
-					Self::swap_currency(who, fee)?;
-					Ok(PaymentSwapResult::Swapped)
+					T::AMMPool::get_spot_price_unchecked(currency, CORE_ASSET_ID, fee)
```
lumir-mrkva, line 356: just note that this price can be wrong or manipulated, it will be safer to use  averaged oracle later and even use fallback price instead if the price from oracle is not based on sufficient liquidity

lumir-mrkva (approved):

pallets/transaction-multi-payment/src/lib.rs
lumir-mrkva: https://github.com/galacticcouncil/Basilisk-node/issues/293

lumir-mrkva:
pallet should be rebenched and benchamrks pallet version bumped

pallets/transaction-multi-payment/src/lib.rs
enthusiastmartin: done

pallets/transaction-multi-payment/src/lib.rs
enthusiastmartin: 👍

Issue #262, opened 2021-11-25
feat: exchange - get AMM pool fee instead of default fee.


enthusiastmartin:
For direct trades - it would be better to get fee defined by the pool it uses.

currently it uses "hard-coded" fee of 0.2%. 

The fee might be different for each pool which is used by the exchange.

PR #291, opened 2021-12-16
feat(exchange): use fee from amm

Roznovjak:
Fixes: #262


enthusiastmartin (changes requested):
Adjust these changes first because AMM does not have generic Fee type anymore.

pallets/xyk/src/lib.rs
```diff
@@ -954,6 +954,11 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
 	fn get_max_out_ratio() -> u128 {
 		T::MaxOutRatio::get()
 	}
+
+	fn get_fee(_pool_account_id: &T::AccountId) -> (u32, u32) {
+		let fee = T::GetExchangeFee::get();
```
vgantchev, line 959: so we are using `Fee::Default()` of 0,2% for any xyk pool? 
just double checking if this is the intention, bc in LBP it is customisable

vgantchev (approved):
lgtm

pallets/xyk/src/lib.rs
Roznovjak: It's customisable in the config, but for all pools.

Roznovjak:
closing this PR and opening new (#302) because this one is more complicated to merge

Issue #286, opened 2021-12-07
Add proxy pallet


jak-pan:
>> https://github.com/galacticcouncil/HydraDX-node/issues/341

lumir-mrkva:
filter has to be defined

PR #292, opened 2021-12-17
feat: add proxy pallet

Roznovjak:
Fixes: #286


Roznovjak:
@jak-pan What extrinsics do we want to allow for proxy accounts?

runtime/common/src/lib.rs
```diff
@@ -111,6 +113,27 @@ parameter_types! {
...
+	Any,
+}
+impl Default for ProxyType {
+	fn default() -> Self {
+		Self::Any
+	}
+}
+
+parameter_types! {
+	pub ProxyDepositBase: Balance = 4 * DOLLARS + 480 * MILLICENTS;
```
vgantchev, line 129: is there some logic behind these amounts?

Roznovjak: This value should be equivalent to what Acala uses.

vgantchev: equivalent, meaning that it takes into account that Snek supply is 100x Acala?

Roznovjak: I used expected dollar values. https://github.com/AcalaNetwork/Acala/blob/ec911c5db90aef7cce00d3143415886369860216/runtime/acala/src/constants.rs#L37

jak-pan: > equivalent, meaning that it takes into account that Snek supply is 100x Acala?

yes dollars mean $ dollar value in sneks

lumir-mrkva (changes requested):
proxy filter has to be defined properly

examples:
kusama - https://github.com/paritytech/polkadot/blob/master/runtime/kusama/src/lib.rs#L1084
polkadot - https://github.com/paritytech/polkadot/blob/master/runtime/polkadot/src/lib.rs#L1082
karura - https://github.com/AcalaNetwork/Acala/blob/master/runtime/karura/src/lib.rs#L1245
acala - https://github.com/AcalaNetwork/Acala/blob/master/runtime/acala/src/lib.rs#L1237

Roznovjak:
not sure if we also want to include `NonTransfer` proxy group or some other use case specific groups.

lumir-mrkva:
add transfer group

Roznovjak:
I have decided to not include cross-chain transfers in the transfer proxy group.

lumir-mrkva (approved):

Issue #261, opened 2021-11-25
refactor: fee calcs using math crate


enthusiastmartin:
Fee calcs have been added to the math crate to make sure that it is calculated in in the same way on FE too.

Use this to calculate fees in XYK and lbp pallets.

PR #294, opened 2021-12-20
refactor: fee calcs from math crate

Roznovjak:
Fixes: #261


enthusiastmartin (commented):
Also exchange and direct trade needs this change.

And the idea is also to completely remove the `WithFee` trait, it is not needed.

pallets/lbp/src/benchmarking.rs
```diff
@@ -36,6 +36,8 @@ const ASSET_B_AMOUNT: Balance = 2_000_000_000;
 const INITIAL_WEIGHT: LBPWeight = 20_000_000;
 const FINAL_WEIGHT: LBPWeight = 90_000_000;
 
+const DEFAULT_FEE: Fee = Fee {numerator: 2, denominator: 1_000};
```
enthusiastmartin, line 39: LBP pallet does not have configurable fee ?

enthusiastmartin (changes requested):
Let's merge first the #296  and reflect the changes in this PR too.

pallets/lbp/src/benchmarking.rs
Roznovjak: Yes, fees are configurable, but not in the config. The fee is set during pool creation.

enthusiastmartin (commented):
let's resolve the build and conflcts (if any with master) and we can merge.

lumir-mrkva (approved):

Issue #306, opened 2022-01-11
Add LBP repay fee to constants


maht0rz:
In order to implement a cleaner client for the node, please add the LBP repay fee (e.g. 20%) to LBP pallet constants.

PR #307, opened 2022-01-11
feat: make repay fee constant

Roznovjak:
Add `repay_fee` into constant metadata.
Fixes: #306


lumir-mrkva (changes requested):
add test and bump version

lumir-mrkva (approved):

PR #308, opened 2022-01-13
feat: add topup auction

green-jay:


pallets/auctions/src/lib.rs
```diff
@@ -596,3 +620,107 @@ impl<T: Config> NftAuction<T::AccountId, T::AuctionId, BalanceOf<T>, Auction<T>>
...
+
+		self.general_data.last_bid = Some((bidder.clone(), value));
+		// Set next minimal bid
+		let bid_step = Permill::from_percent(<T as crate::Config>::BidStepPerc::get()).mul_floor(value);
+		self.general_data.next_bid_min = value.checked_add(&bid_step).ok_or(Error::<T>::BidOverflow)?;
+
+		let current_block = <frame_system::Pallet<T>>::block_number();
+
+		self.specific_data.bids.push(Bid {
+			who: bidder,
```
vgantchev, line 647: mayb `bidder` instead of who?

pallets/auctions/src/lib.rs
```diff
@@ -596,3 +620,107 @@ impl<T: Config> NftAuction<T::AccountId, T::AuctionId, BalanceOf<T>, Auction<T>>
...
+			when: current_block,
```
vgantchev, line 648: `timestamp` or `block_nr`? not rly sure about the conventions :)

green-jay:
> The main design flaw I see here is that the starting bid must always be above the minimum price (next_bid_min in `validate_bid()`). As this is a charity auction, we should probably allow people to place bids below the starting price. Eventually, even smaller bids on an expensive NFT will add up.
> 
> To achieve this, I suggest introducing `general_data.reserve_price`. In the case of English, `reserve_price` will be set to `next_bid_min`, so the first bid must always meet the reserve price (bots)
> 
> In the case of topup, we can accept bids below the reserve_price and use next_bid_min for the bid step calculation. Once the auction closes, we can transfer the frozen funds if the aggregated bids >= reserve price, otherwise unfreeze the funds
> 
> wdyt?

Generally introducing a reserve price is a good idea and could be a nice stimulus for user activity. On the other hand lacking a reserve price is definitely not a design flaw. I think of it as an additional feature we can easily implement. I'll create an issue for later so this doesn't get lost

pallets/auctions/src/lib.rs
green-jay: definitely not timestamp :) i think i might remove it completely, not useful for TopUp, will be used in candle i suppose

vgantchev:
Another missing feature is beneficiary_address - since this is for charity, we could provide the option to set a different address which will be receive the funds. This address will be publicly visible, and will help to a certain extent in preventing the risk of the seller keeping the funds for himself.

But this could also be an issue for future implementation, if you prefer it that way.

I still need to check tests before final review

green-jay:
> Another missing feature is beneficiary_address - since this is for charity, we could provide the option to set a different address which will be receive the funds. This address will be publicly visible, and will help to a certain extent in preventing the risk of the seller keeping the funds for himself.
> 
> But this could also be an issue for future implementation, if you prefer it that way.
> 
> I still need to check tests before final review

yes, makes absolute sense. please add that to the issue I created (or create a new one)

PR #312, opened 2022-01-17
feat: tx multi payment  - revised spot price approach

enthusiastmartin:
This PR covers the change to use XYK spot price in the multi payment pallet. 

Includes following refactors:
- moved multi payment pallet to warehouse
- spot price provider by XYK is done in generic way, so it can be easily swapped for eg. price-oracle when needed
- spot prices are "loaded" on init so it does not need to fetch spot price each time payment is required.
- removed  additional base weight ( currently using substrate base weight - pending custom base weight benchmark - will be done as separate PR )


pallets/xyk/src/impls.rs
```diff
@@ -0,0 +1,31 @@
...
+impl<T: crate::Config > SpotPriceProvider<AssetId> for XYKSpotPrice<T>{
+    type Price = Price;
+
+    fn pair_exists(asset_a: AssetId, asset_b: AssetId) -> bool {
+        <crate::Pallet<T>>::exists(AssetPair::new(asset_b,asset_a))
+    }
+
+    fn spot_price(asset_a: AssetId, asset_b: AssetId) -> Option<Self::Price> {
+        if Self::pair_exists(asset_a, asset_b) {
+            let pair_account = <crate::Pallet<T>>::get_pair_id(AssetPair {
```
lumir-mrkva, line 19: this branch could be extracted into overidable function as rest of it will work for all cases probably

lumir-mrkva (changes requested):
bump versions pls

lumir-mrkva (approved):

PR #354, opened 2022-02-03
ci: change instance type from c5ad to c5a

fakirAyoub:


martinfridrich (approved):

lumir-mrkva:
@fakirAyoub whats the reason for the change, the difference in the machine? can we get some faster one still?

PR #360, opened 2022-02-05
chore: weights update 

enthusiastmartin:
This PR includes:

- update of pallet weights
- Extrinsic Base Weight set to 200 * WEIGHTS_PER_MICROS which includes the payment in non-native currency
- weights moved into common runtime so the same weights are used in both runtimes and it is easy to update/maintain


Roznovjak (approved):

Roznovjak:
please bump runtime versions

lumir-mrkva (approved):

PR #363, opened 2022-02-07
refactor: pin runtime pallet ids

### LINKED ISSUE #346: Persistent id's in Runtime enum
this will allow to add/remove pallets without changing id of the pallet on which consumers rely

enthusiastmartin:
PR makes sure that each pallet in runtime is assigned persistent id. 

This makes sure that module ids are the same when a pallet is either removed or added. 

I decided to group ids as follows:
 - starting at 0 - substrate related pallets
 - starting at 50 - parachain and xcm related stuff
 - stating at 100 - basilisk pallets and warehouse pallets
 - starting at 150 - orml
 - starting at 200 - testing stuff - pallets which are not present in basilisk, only in testing runtime

The reasons being:
 - allows quicker identify area of an error
 - allows adding new pallets to be together with related stuff
 - order is also important - therefore having some spaces prevents unwanted shifts in future. 

Note that the order of pallets is important. 

Eg. 
- Session must be after collator selection and prior to aura. 
- Asset registry must be before orml-tokens otherwise it can cause issue with existential deposit.

fixes #346

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -503,7 +503,7 @@ dependencies = [
 
 [[package]]
 name = "basilisk-runtime"
-version = "35.0.0"
+version = "36.0.0"
 dependencies = [
  "common-runtime",
  "cumulus-pallet-aura-ext",
@@ -10926,7 +10926,7 @@ dependencies = [
 
 [[package]]
 name = "testing-basilisk-runtime"
-version = "35.0.0"
+version = "36.0.0"
 dependencies = [
  "common-runtime",
  "cumulus-pallet-aura-ext",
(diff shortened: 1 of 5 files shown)


Roznovjak:
What about moving testing stuff to 250 or 300 and leaving some space for potential new groups? It's just an idea.

Roznovjak (approved):

enthusiastmartin:
> What about moving testing stuff to 250 or 300 and leaving some space for potential new groups? It's just an idea.

it is u8. max is 255.

lumir-mrkva (approved):

PR #369, opened 2022-02-08
ci: fixed runner destroy

fakirAyoub:
Fixes the bug that was causing runners not to be destroyed during the destroy step

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -149,7 +149,7 @@ jobs:
   destroy-runner-instance:
     name: destroy runner
     runs-on: ubuntu-latest
-    needs: pop-runner-instance
+    needs: [ pop-runner-instance, build, build-and-publish-docker-image]
     if: ${{ always() && github.event.pull_request.head.repo.fork == false }}
     env:
       ACTIONS_ALLOW_UNSECURE_COMMANDS: true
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yml
```diff
@@ -22,17 +21,20 @@ jobs:
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       AWS_REGION: ${{ secrets.AWS_REGION }}
       EC2_SECRET: ${{ secrets.EC2_PWD }}
-      S3_BACKEND_KEY_NAME: tf-state-${{ github.sha }}
+      S3_BACKEND_KEY_NAME: TMP
     outputs:
-      runner: ${{ steps.determine-runner.outputs.runner }}
+      runner: ${{ env.S3_BACKEND_KEY_NAME }}
```
lumir-mrkva, line 26: "TMP" will be output, you have to specify which job you want to take output from

.github/workflows/workflow.yml
```diff
@@ -144,19 +146,18 @@ jobs:
           path: .
       - name: S3 Upload basilisk-testnet-chainspec.json
         uses: tpaschalis/s3-sync-action@master
-
   destroy-runner-instance:
     name: destroy runner
     runs-on: ubuntu-latest
-    needs: [pop-runner-instance, build, build-and-publish-docker-image]
+    needs: pop-runner-instance
```
lumir-mrkva: this will imidiatelly destroy runner after creating it, not sure if we want that

.github/workflows/workflow.yml
fakirAyoub: The value is erased during the action

fakirAyoub: `run: echo "S3_BACKEND_KEY_NAME=$(echo $(uuidgen))" >> $GITHUB_ENV`

lumir-mrkva: also this breaks behaviour in forks

lumir-mrkva (approved):

PR #371, opened 2022-02-09
feat: moonbase & rococo chainspec

lumir-mrkva:


node/src/testing_chain_spec.rs
```diff
@@ -291,6 +297,84 @@ pub fn k8s_testnet_parachain_config() -> Result<ChainSpec, String> {
...
+		// Protocol ID
+		Some(PROTOCOL_ID),
+		// Fork ID
+		None,
+		// Properties
+		Some(properties),
+		// Extensions
+		Extensions {
+			relay_chain: "westend".into(),
+			para_id: PARA_ID,
```
martinfridrich, line 373: Can we register with arbitrary parachain id? This will use `2090`

martinfridrich (commented):
I wasn't able to run it, look at my commit.

node/src/testing_chain_spec.rs
jak-pan: We can set different constant if needed I guess?

martinfridrich: yes, I realized now this para id must be different otherwise it will collide with basilisk on testnet.

lumir-mrkva: on moonvase we can, for rococo we have different id

PR #372, opened 2022-02-09
refactor: rename run_to_block to set_block_number

### LINKED ISSUE #317: Rename run_to_block  in LBP
fn run_to_block(n) used in LBP tests and some other pallets and it is not doing what I would expect = iterating in a loop across all blocks until block n and initialize/finalize them. It is just setting block to n directly thus should be renamed accordingly (to set_block_number or alike)

vgantchev:
Resolves #317

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5809,7 +5809,7 @@ dependencies = [
 
 [[package]]
 name = "pallet-lbp"
-version = "4.3.1"
+version = "4.3.2"
 dependencies = [
  "frame-benchmarking",
  "frame-support",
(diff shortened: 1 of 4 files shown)


vgantchev:
It is complaining about version check - is version bump needed given that only mock and tests were updated?

lumir-mrkva:
> It is complaining about version check - is version bump needed given that only mock and tests were updated?

yes, currently any change in the src has to be versioned

jak-pan:
LGTM after bump. Do we want to bump if only tests change tho? If not then we should have issue for fixing this.

green-jay (approved):

PR #384, opened 2022-02-15
fix: non native fee payment spot price

### LINKED ISSUE #330: integration test for tx fee in multi-payment scenario
There was a change in multi payment pallet recently where if tx fee is paid in non-native currency - the spot price for given asset pair is provided by xyk

Perform a test to verify that it is all calculated and used correctly. This would involve following steps:

1. set tx fee currency to non-native asset.
2. Make some transactions to verify if tx fee is paid in that asset and the set  fallback price is correctly used and correct amount is withdrawn
3. create a pool between native asset and the asset ☝️ 
4. optionally perform some trades so the price changes ( bonus: check price-oracle price )
5. repeat step 5 and verify that the correct spot price from xyk pool is used ( there should be event like FeeWithdrawn )
6. write an integration test for this scenario so any changes in this area would capture any discrepancies.

### LINKED ISSUE #386: bug: fix spot price - asset order is incorrect
Spot price provider in XYK calculates the spot price as asset A / asset B. This is not correct and should be asset B / asset A instead.

Example: 
A = BTC (price=40000)
B = ETH (price=2000)

Spot price of BTC/ETH is currently 20 but should be 0.05 (aka what number do I need to multiply the price of asset A to get to price of asset B)

This should be reflected in Fee calculation which accounts for the reverted price and calculates the Fee correctly atm.

green-jay:
fixes #330 
 fixes #386


green-jay:
@lumir-mrkva this issue is mainly about adding a new integration test, hence "test"

runtime/basilisk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "basilisk-runtime"
-version = "38.0.0"
+version = "39.0.0"
```
enthusiastmartin, line 3: @lumir-mrkva  well, runtime change again. 

in this case - is it really  needed here ?  there is only one change and it only affects benchmarking.

lumir-mrkva: critical bug in xyk was fixed

lumir-mrkva (approved):

PR #385, opened 2022-02-15
refactor: TopUp Auction

vgantchev:
This PR builds on https://github.com/galacticcouncil/Basilisk-node/pull/308
It delivers the missing pieces for finishing TopUp auctions

When a user bids on a topup auction, the amount which is bid is reserved (ie moved to a subaccount belonging to the auction). 

At auction close, the whole amount is transferred to the seller in case the lot is sold (ie reserve_price is met)

If there is no winner because the reserve price was not met, all bidders can claim back their (aggregated) reserved amounts

TODO: 
* Add documentation


green-jay:
some of the docs are obsolete so that needs revisiting but LGTM

Issue #380, opened 2022-02-15
Generate rustdocs workflow


lumir-mrkva:
- [ ] generate runtime / rustdocs
- [ ] deploy docs to gh pages
- [ ] set custom domain
- [ ] link to https://docs.bsx.fi/
- [ ] add to ci workflow

Roznovjak:
update on the workflow I am testing:

- [ ] create `gh-pages` branch where the docs will be stored
- [ ] set the `gh-pages` branch for github pages in the Settings -> Pages -> Source
- [ ] the docs are built conditionally when a PR is merged into the master branch to avoid regenerating and deploying of the docs on every commit to a PR
- [ ] build the docs in the existing workflow along the build process to avoid rebuilding the sources
- [ ] force push the docs to the `gh-pages` branch

- [ ] create dummy `index.html` file in the root dir of the `gh-pages` branch redirecting to the main `index.html` file nested in a subfolder. Otherwise the gh page results in the 404 error.
- [ ] add link to the docs from `docs.bsx.fi`

Roznovjak:
Had to create a separate workflow. It looks like it is not possible to conditionally run a job from our existing workflow only when a PR is merged.

lumir-mrkva:
> Had to create a separate workflow. It looks like it is not possible to conditionally run a job from our existing workflow only when a PR is merged.

separate workflow can be ok in this case, but regardless, this should be possible, we are already doing something similar for testnet deployment https://github.com/galacticcouncil/Basilisk-node/blob/master/.github/workflows/workflow.yml#L193
(check that this is not running in PR and only in correct branch)

Roznovjak:
> > Had to create a separate workflow. It looks like it is not possible to conditionally run a job from our existing workflow only when a PR is merged.
> 
> separate workflow can be ok in this case, but regardless, this should be possible, we are already doing something similar for testnet deployment https://github.com/galacticcouncil/Basilisk-node/blob/master/.github/workflows/workflow.yml#L193 (check that this is not running in PR and only in correct branch)

I tried to use this condition `if: github.event.pull_request.merged == true`, but it was not set in our workflow.

PR #389, opened 2022-02-21
ci: generate and deploy rust docs

Roznovjak:
Fixes: #380

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,14 +1,12 @@
 name: Deploy docs to GitHub pages
 
 on:
-  pull_request:
-    types: [closed]
+  push:
     branches:
       - master
       - generate_docs
 jobs:
   generate-and-deploy-docs:
-    if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,11 +1,18 @@
 name: Deploy docs to GitHub pages
 
 on:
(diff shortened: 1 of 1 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,31 @@
+name: Deploy docs to GitHub pages
+
+on:
+  pull_request:
```
lumir-mrkva, line 4: i would only run it in branch master and this branch (for testing)

this would also rewrite docs in case we merge pr to some other branch

lumir-mrkva (changes requested):

.github/workflows/docs.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+      - name: Generate docs
+        run: |
+          cargo doc --workspace --all-features --release --no-deps --document-private-items --target-dir ./generated_docs
+          cd generated_docs/doc
+          echo '<meta http-equiv="refresh" content="0; url=https://galacticcouncil.github.io/Basilisk-node/basilisk/index.html">' > index.html
```
lumir-mrkva, line 19: redirect should be relative, not absolute

have you tried to just deploy `basilisk` subdirectory? that would remove need for redirect all together

Roznovjak: `basilisk` subdirectory is just for the documentation of the basilisk binary. All subdirectories are required. Each target has its own subdirectory.

lumir-mrkva (approved):

PR #390, opened 2022-02-21
feat: add candle auction

green-jay:


green-jay:
upgrade to master and providing randomness done, merging for better rebasing

PR #394, opened 2022-02-23
feat: Candle Auction

### LINKED ISSUE #388: Add Candle auction

vgantchev:
Resolves #388 

Candle auction + refactor

TODO: docs


green-jay (commented):
review improvements/suggestions: https://github.com/galacticcouncil/Basilisk-node/pull/400

Storage HighestBiddersByAuctionClosingRange should not only bloat but get removed/pruned at some point after closing.

Other than that - lgtm

green-jay:
Also thinking we might want to split the test file to multiple, maybe one per auction, so we don't get such a large test file

PR #395, opened 2022-02-24
ci: trigger workflow for rust docs only on merge

Roznovjak:


lumir-mrkva:
@Roznovjak we already only run it on merge, as there shouldn't be anything pushed into master otherwise. or i am missing something?

Roznovjak:
> @Roznovjak we already only run it on merge, as there shouldn't be anything pushed into master otherwise. or i am missing something?

Yes, you are right. I didn't realize that a combination of `push` + `branches:master` has similar effect as the change in this PR.

PR #405, opened 2022-03-07
refactor: Split auction tests

vgantchev:
This PR splits the various auctions tests in modules


martinfridrich:
I don't think `mod {...}` is necessary in `*_test.rs` files and I would remove the `_tests` suffix. I think `_tests` is redundant, files are in the `tests` dir.

Issue #327, opened 2022-01-25
Test code formating in CI


lumir-mrkva:
make sure that PR has correctly formated code before merge

PR #422, opened 2022-04-07
ci: check formatting

Roznovjak:
Fixes: https://github.com/galacticcouncil/warehouse/issues/41
Fixes: #327


.github/workflows/workflow.yml
```diff
@@ -86,6 +86,8 @@ jobs:
         uses: codota/toolchain@00a8bf2bdcfe93aefd70422d3dec07337959d3a4
         with:
           profile: minimal
+      - name: Check formatting
```
lumir-mrkva, line 89: i would move it to a separate job, mby even separate workflow altogether as you may want to see results of the build regardless of your formatting

lumir-mrkva (approved):

PR #423, opened 2022-04-08
feat!: remove faucet in testing runtime

jak-pan:
This brings a question if to bump only testing runtime or keep consistent?
- funded drip account for kubernetes testnet
- removed faucet


lumir-mrkva (changes requested):
you have to increase version of client as well when you change chainspec and non testing runtime as well to keep them in sync

also you have to update lockfile after version bump

Roznovjak (approved):

lumir-mrkva:
removal of pallet from chain spec probably isn't worth of major version change

jak-pan:
> removal of pallet from chain spec probably isn't worth of major version change

It is a breaking change.

lumir-mrkva:
> > removal of pallet from chain spec probably isn't worth of major version change
> 
> It is a breaking change.

of the runtime, not the client node

lumir-mrkva (approved):

Issue #334, opened 2022-01-25
XCM: add deposit error handler for multi-currency adapter


enthusiastmartin:
more info -> https://github.com/open-web3-stack/open-runtime-module-library/pull/675

_Originally posted by @lumir-mrkva in https://github.com/galacticcouncil/Basilisk-node/pull/328#discussion_r791593408_

lumir-mrkva:
currently the default implementation is used

enthusiastmartin:
I think we should use `DepositToAlternative `implementation.

this will deposit to selected alternative account ( eg. treasury ) instead of failure. 

@lumir-mrkva what you think ?

lumir-mrkva:
i agree that is better to deposit token somewhere than just ignoring it, but i would create some special account for this use case

PR #431, opened 2022-04-14
feat: add XCM deposit event handler

Roznovjak:
Fixes: #334


runtime/basilisk/src/xcm.rs
```diff
@@ -269,6 +270,10 @@ pub type LocationToAccountId = (
 	AccountId32Aliases<RelayNetwork, AccountId>,
 );
 
+parameter_types! {
+	pub Alternative: AccountId = PalletId(*b"xcm/alte").into_account();
```
enthusiastmartin, line 274: i would add some brief description what is this account for.

enthusiastmartin (changes requested):
Probably testing runtime needs the same change.

lumir-mrkva:
bump runtime versions and LGTM

PR #433, opened 2022-04-22
build: use asset-registry from warehouse

Roznovjak:


lumir-mrkva (approved):
LGTM

lumir-mrkva:
update missing versions and merge it ;)

PR #442, opened 2022-05-03
fix: add missing pallet-proxy in std feature

apopiak:
`cargo check -p basilisk-runtime` was failing when trying to compile `pallet-proxy` and I noticed it was missing in the std feature list for the runtime.

### DIFF
--- a/runtime/basilisk/Cargo.toml
+++ b/runtime/basilisk/Cargo.toml
@@ -221,6 +221,7 @@ std = [
     "pallet-treasury/std",
     "pallet-collective/std",
     "pallet-democracy/std",
+    "pallet-proxy/std",
     "pallet-scheduler/std",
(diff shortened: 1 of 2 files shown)


enthusiastmartin (approved):

apopiak:
should I bump impl_version? shouldn't need a spec_version bump AFAICT

lumir-mrkva:
> should I bump impl_version? shouldn't need a spec_version bump AFAICT

yes, we use simple runtime versioning, any PR that changes runtime in any way has to increment spec version to ensure that we dont have 2 different runtimes with same spec

apopiak:
so I will bump spec, then

lumir-mrkva:
> so I will bump spec, then

we also keep the cargo version in sync with spec version and testing runtime spec version as well

PR #443, opened 2022-05-03
feat!: XCM Weight Trader

### LINKED ISSUE #373: xcm: implement trader

apopiak:
## Description
The weight trader uses `pallet-transaction-multi-payment` to determine whether the payment asset has a local price (and can be used for fee payment).
It then uses `WeightToFee` in combination with that price to calculate the fee to charge for the weight.
The trader tracks assets used for payment and uses them naively (i.e. one by one) to refund weight.

## Related
closes #373 

based on https://github.com/galacticcouncil/warehouse/pull/48

## Motivation and Context
Basilisk currently does not charge for XCM execution which makes it open for exploitation.

## How Has This Been Tested?

+ Unit tests in https://github.com/galacticcouncil/warehouse/pull/48
+ Adjusted integration tests to make sure that the treasury receives fees on XCM transfer.

## Checklist:


apopiak:
### Todos / Open issues
- [ ] This should be in the warehouse so it can be used by the HydraDx runtime.
  - [ ] move `PriceOracle` trait
  - [ ] move `MultiCurrencyTrader`
- [ ] Should this be using the regular `WeightToFee` from the runtime or something that makes for cheaper execution (@lumir-mrkva 's suggestion)?

apopiak:
>   Should this be using the regular `WeightToFee` from the runtime or something that makes for cheaper execution (@lumir-mrkva 's suggestion)?

It will use regular `WeightToFee` for now. We can optimize fees later if it's necessary.

runtime/basilisk/src/xcm.rs
```diff
@@ -96,7 +99,25 @@ impl Config for XcmConfig {
 
 	type Barrier = Barrier;
 	type Weigher = FixedWeightBounds<BaseXcmWeight, Call, MaxInstructions>;
-	type Trader = TradePassthrough;
+	// We calculate weight fees the same way as for regular extrinsics and use the prices and choice
+	// of accepted currencies of the transaction payment pallet. Burn fee revenue.
```
apopiak, line 103:
```suggestion
	// We calculate weight fees the same way as for regular extrinsics and use the prices and choice
	// of accepted currencies of the transaction payment pallet. Fees go to the same fee receiver as
	// configured in `MultiTransactionPayment`.
```

enthusiastmartin (approved):

Issue #427, opened 2022-04-14
Add check if nft class id is from reserved range - liquidity mining


martinfridrich:
https://github.com/galacticcouncil/Basilisk-node/pull/250#issuecomment-1087518332

PR #444, opened 2022-05-05
feat: ensure liquidity mining NFT id is within reserved range

Roznovjak:
Fixes: #427

### DIFF
--- a/pallets/liquidity-mining/Cargo.toml
+++ b/pallets/liquidity-mining/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-liquidity-mining"
-version = "1.0.1"
+version = "1.0.2"
 description = "Liquidity mining"
 authors = ["GalacticCouncil"]
 edition = "2018"
--- a/pallets/liquidity-mining/Cargo.toml
+++ b/pallets/liquidity-mining/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-liquidity-mining"
-version = "1.0.2"
+version = "1.1.0"
 description = "Liquidity mining"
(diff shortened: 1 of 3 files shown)


pallets/liquidity-mining/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "pallet-liquidity-mining"
-version = "1.0.1"
+version = "1.0.2"
```
lumir-mrkva, line 3: as this change adds another feature, i would increase version to `1.1.0`

lumir-mrkva:
@Roznovjak can you add test that check works like expected?

Roznovjak:
> @Roznovjak can you add test that check works like expected?

A test is automatically generated from it. But it's more like a verification, because testing it (both scenarios) would require two runtimes with different pallet configurations.

lumir-mrkva:
> > @Roznovjak can you add test that check works like expected?
> 
> A test is automatically generated from it. But it's more like a verification, because testing it (both scenarios) would require two runtimes with different pallet configurations.

when it actually panics? after runtime upgrade?

Roznovjak:
> > > @Roznovjak can you add test that check works like expected?
> > 
> > 
> > A test is automatically generated from it. But it's more like a verification, because testing it (both scenarios) would require two runtimes with different pallet configurations.
> 
> when it actually panics? after runtime upgrade?

It looks like it behaves just like a normal test. [Ref](https://github.com/paritytech/substrate/pull/6356)

lumir-mrkva (approved):
found the test: `test mock::__construct_runtime_integrity_test::runtime_integrity_tests ... ok`

LGTM

Issue #206, opened 2021-10-20
ci: run benchmarks in test to verify they can be executed


mrq1911:
passing test benchmarks in mocked environment doesn't guarantee that they will work in the runtime

and if we're building benchmarking binaries, we can release them as well https://github.com/galacticcouncil/HydraDX-node/issues/313

PR #452, opened 2022-05-18
refactor: use runtime env in benchmark tests

Roznovjak:
Related to #206


enthusiastmartin (commented):
personally, i dont see the benefit of this change.

benchmarks are always run in selected environment. 

What was exactly the motivation ?

Roznovjak:
> personally, i dont see the benefit of this change.
> 
> benchmarks are always run in selected environment.
> 
> What was exactly the motivation ?

The main reason for this is probably that running benchmarks in selected environments doesn't guarantee that benchmarks can be successfully executed. We have had this problem at least once. Actually, all the tests pass in the master branch, but we can't benchmark all the pallets we use. Tests guarantee only that benchmarks can be executed in the same environment. 
@lumir-mrkva maybe knows some other reasons.

lumir-mrkva:
> > personally, i dont see the benefit of this change.
> > benchmarks are always run in selected environment.
> > What was exactly the motivation ?
> 
> The main reason for this is probably that running benchmarks in selected environments doesn't guarantee that benchmarks can be successfully executed. We have had this problem at least once. Actually, all the tests pass in the master branch, but we can't benchmark all the pallets we use. Tests guarantee only that benchmarks can be executed in the same environment. @lumir-mrkva maybe knows some other reasons.

i agree with @Roznovjak about need to verify that benchmarks can be actually ran, i am just not sure what is the purpose of this PR vs #453

Roznovjak:
> > > personally, i dont see the benefit of this change.
> > > benchmarks are always run in selected environment.
> > > What was exactly the motivation ?
> > 
> > 
> > The main reason for this is probably that running benchmarks in selected environments doesn't guarantee that benchmarks can be successfully executed. We have had this problem at least once. Actually, all the tests pass in the master branch, but we can't benchmark all the pallets we use. Tests guarantee only that benchmarks can be executed in the same environment. @lumir-mrkva maybe knows some other reasons.
> 
> i agree with @Roznovjak about need to verify that benchmarks can be actually ran, i am just not sure what is the purpose of this PR vs #453

#453 is related to the second part of the task (2nd line in the description). There is a link to [this](https://github.com/galacticcouncil/HydraDX-node/issues/313) issue.

enthusiastmartin:
You don't need to do this - delete benchmarks from a pallet and move it to runtime. 

If you run a benchmark of selected pallet. It will run with selected runtime. 

That means this change much.

So there is no need to do this.

I believe it is better to keep benchmarks as part of the pallet if possible. Usually it is possible, there are currently only few exceptions where it is need to be part of runtime ( eg orml). 

 If you want to run benchmarks in ci - the other pr is sufficient.

Roznovjak:
> You don't need to do this - delete benchmarks from a pallet and move it to runtime.
> 
> If you run a benchmark of selected pallet. It will run with selected runtime.
> 
> That means this doesn't change much.
> 
> So there is no need to do this.
> 
> I believe it is better to keep benchmarks as part of the pallet if possible. Usually it is possible, there are currently only few exceptions where it is need to be part of runtime ( eg orml).
> 
> If you want to run benchmarks in ci - the other pr is sufficient.

The main goal of this task is to execute benchmarking tests in the main runtime and not in mocked environments.

Roznovjak:
> > 
> 
> i see.
> 
> i still don't like the idea of separating the benchmarks from the pallet. Because if you change something in a pallet, you will need to go and l look for this benchmarks in a runtime.
> 
> you should be able to use the runtime to ruin the tests without the need of moving the sources into the runtime ( providing that the benchmarks are publicly visible ).
> 
> I havent tried but i currently dont see why it should not work ( but ofc, i might be wrong )

It looks like benchmarking tests are not publicly available.

lumir-mrkva:
stale

PR #453, opened 2022-05-18
ci: build, test and publish benchmarking binary

Roznovjak:
Fixes: #206


.github/workflows/workflow.yml
```diff
@@ -126,6 +126,17 @@ jobs:
         with:
           name: basilisk_runtime.compact.compressed.wasm
           path: target/release/wbuild/basilisk-runtime/basilisk_runtime.compact.compressed.wasm
+      - name: Build benchmarking node
+        run: cargo build --release --locked --features runtime-benchmarks
+      - name: Version info
+        run: ./target/release/basilisk --version
+      - name: Run benchmarktests
```
lumir-mrkva, line 133: i think that these tests are already ran with coverage tests, but this issue is mainly about executing benchmarks itself from binary, not tests if them

lumir-mrkva (changes requested):

.github/workflows/workflow.yml
Roznovjak: Updated. But benchmarks fail because there is a problem (session keys are not registered for validators used in the benchmark) with the collator-selection pallet (in cumulus v0.9.17). Upgrading will fix this issue. I also removed the tips pallet because we use the default weights for it and can't be benchmarked if `elections-phragmen::DesiredMembers < 2` (and is set to 1).

enthusiastmartin:
Personally, i would not add this as part of every CI run.

And as it is - i don't see much benefit of it. The results are not published, benchmarks will never fail ( unless problem with compilation) so i doubt anyone is going to look at it anyway.

Benchmarks tests are already executed as part of test target.

we already build everything multiple times ( clippy, tests, node) and this will make things taking even longer. 

if this is really needed, i would do it only on PR merge to master. 

i believe the benchmarking target is not going to be used. When we do update weights, we just build it on reference machine anyway.

lumir-mrkva:
> Personally, i would not add this as part of every CI run.
> 
> And as it is - i don't see much benefit of it. The results are not published, benchmarks will never fail ( unless problem with compilation) so i doubt anyone is going to look at it anyway.
> 
> Benchmarks tests are already executed as part of test target.
> 
> we already build everything multiple times ( clippy, tests, node) and this will make things taking even longer.
> 
> if this is really needed, i would do it only on PR merge to master.
> 
> i believe the benchmarking target is not going to be used. When we do update weights, we just build it on reference machine anyway.

i don't agree, this is done specifically so we check whenever we can really run the benchmarks, because it happened many times in past situation where benchmark test passed, but benchmark run failed

Roznovjak:
> benchmark run has to succeed

Imho failing benchmarks should be fixed in a different task.

enthusiastmartin:
> > Personally, i would not add this as part of every CI run.
> > 
> > And as it is - i don't see much benefit of it. The results are not published, benchmarks will never fail ( unless problem with compilation) so i doubt anyone is going to look at it anyway.
> > 
> > Benchmarks tests are already executed as part of test target.
> > 
> > we already build everything multiple times ( clippy, tests, node) and this will make things taking even longer.
> > 
> > if this is really needed, i would do it only on PR merge to master.
> > 
> > i believe the benchmarking target is not going to be used. When we do update weights, we just build it on reference machine anyway.
> 
> i don't agree, this is done specifically so we check whenever we can really run the benchmarks, because it happened many times in past situation where benchmark test passed, but benchmark run failed

Well.   Fine by me.

Can we run it in parallel at least?

lumir-mrkva:
> > benchmark run has to succeed
> 
> Imho failing benchmarks should be fixed in a different task.

doesn't make difference to me whenever this is fixed in this PR or some other one, but this PR cannot be merged failing

.github/workflows/workflow.yml
```diff
@@ -127,6 +127,31 @@ jobs:
           name: basilisk_runtime.compact.compressed.wasm
           path: target/release/wbuild/basilisk-runtime/basilisk_runtime.compact.compressed.wasm
 
+  build-benchmarks:
+    needs: pop-runner-instance
+    runs-on: ${{ needs.pop-runner-instance.outputs.runner }}
```
lumir-mrkva, line 132: this is not running in parallel, it just waits for runner to be free. you'll have to set it to github runner "ubuntu-latest" and you can remove need for pop-runner as you are nt using that runner anymore

lumir-mrkva (changes requested):

Roznovjak:
> > > benchmark run has to succeed
> > 
> > 
> > Imho failing benchmarks should be fixed in a different task.
> 
> doesn't make difference to me whenever this is fixed in this PR or some other one, but this PR cannot be merged failing

Ok. In this case we need to wait for upgrade of polkadot dependencies. It will fix the failing benchmark.

enthusiastmartin:
Just wondering.. Why do we need to run benchmarks ? Unless we do something
with the results themselves, I don't see any benefit to run them.

Benchmarks tests should be executed as part of test target.

lumir-mrkva:
stale

PR #455, opened 2022-05-19
fix!: use same version of orml-utilities

apopiak:
the marketplace pallet was using a different version of orml-utilities (polkadot-v0.9.16) which can lead to build errors

## Checklist:


lumir-mrkva (changes requested):
pls push lockfile and bump marketplace version

lumir-mrkva (approved):

PR #456, opened 2022-05-21
feat: Add benchmarks for auctions

vgantchev:
This PR:

* adds benchmarks and weights for all existing auction types. For now, the difference in the weights is not so big as to require refund of fees. One exception is candle.close but this will change.
* adds an abstraction for mocked objects which can be used both in testing and benchmark runtimes (`mocked_auction_objects.rs`)
* addresses most of the review comments on #168

I couldn't find a way to mock the randomness to generate the worst-case scenario for candle.close() which is 1 bid place in range 1 (this would cause 100+ reads). In any case, the storage will be changed following the suggestions of @apopiak


pallets/auctions/src/lib.rs
```diff
@@ -363,8 +366,13 @@ pub mod pallet {
...
 		///
 		/// - calls the create() implementation on the given Auction type
-		/// 
-		#[pallet::weight(<T as Config>::WeightInfo::create_auction())]
+		///
+		#[pallet::weight(
+			<T as Config>::WeightInfo::create_english()
+				.max(<T as Config>::WeightInfo::create_topup())
+				.max(<T as Config>::WeightInfo::create_candle())
+		)]
```
enthusiastmartin, line 374: i would probably go with parametrize create_auction weight function where the create_auction would take a type of auction. 

In that case, you could actually return correct weight and not the max weight .

same could be done for all other extrinsics?

PR #467, opened 2022-06-01
test: add some invariants tests for xyk pallet

enthusiastmartin:
Similar to what was done for other amm pool implementation - let's add some property based invariants tests for xyk too.

### DIFF
--- a/pallets/xyk/src/invariants.rs
+++ b/pallets/xyk/src/invariants.rs
@@ -1,0 +1,121 @@
+use primitive_types::U256;
+use primitives::Balance;
+use proptest::prelude::*;
+use sp_runtime::FixedU128;
+
+pub const ONE: Balance = 1_000_000_000_000;
+const TOLERANCE: Balance = 1_000;
+
+#[macro_export]
+macro_rules! assert_eq_approx {
+	( $x:expr, $y:expr, $z:expr, $r:expr) => {{
+		let diff = if $x >= $y { $x - $y } else { $y - $x };
+		if diff > $z {
+			panic!("\n{} not equal\n left: {:?}\nright: {:?}\n", $r, $x, $y);
+		}
+	}};
+}
+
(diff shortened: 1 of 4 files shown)


pallets/xyk/src/invariants.rs
```diff
@@ -0,0 +1,121 @@
...
+
+		let p0 = FixedU128::from((asset_a_reserve, asset_b_reserve));
+		let p1 = FixedU128::from((asset_a_reserve + amount, asset_b_reserve + amount_b));
+
+		// Price should not change
+		assert_eq_approx!(p0,
+			p1,
+			FixedU128::from_float(0.0000000001),
+			"Price has changed after add liquidity");
+	}
```
poliwop, line 99: Add a check that delta_S * R <= delta_R * S, with *no* tolerance... this ensures that we are rounding such that we do not mint too many shares in a way that could lead to exploit, on liquidity add.

pallets/xyk/src/invariants.rs
```diff
@@ -0,0 +1,121 @@
...
+
+		let p0 = FixedU128::from((asset_a_reserve, asset_b_reserve));
+		let p1 = FixedU128::from((asset_a_reserve - amount_a, asset_b_reserve - amount_b));
+
+		// Price should not change
+		assert_eq_approx!(p0,
+			p1,
+			FixedU128::from_float(0.0000000001),
+			"Price has changed after add liquidity");
+	}
```
poliwop, line 120: Add a check that delta_S * R >= delta_R * S, with *no* tolerance... this ensures that we are rounding such that we do not give out too many assets in a way that could lead to exploit, on liquidity remove.

poliwop (commented):
Tests look fine, but I think these tests probably belong in hydra_dx_math since they are testing hydra_dx_math functions? We can (and should, with negotiable priority) add similar invariant tests for xyk pallet functions validate_buy, validate_sell, add_liquidity and remove_liquidity.

PR #475, opened 2022-06-07
chore!: refer to warehouse deps with one slash everywhere

apopiak:
found more double slash deps


enthusiastmartin (approved):
I also double-checked other patch github links (cumulus,polkadot,substrate) to make sure no double slashes used.

runtime/basilisk/src/lib.rs
```diff
@@ -105,7 +105,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("basilisk"),
 	impl_name: create_runtime_str!("basilisk"),
 	authoring_version: 1,
-	spec_version: 48,
+	spec_version: 49,
```
enthusiastmartin, line 108: i dont think this is needed.

version check complains. but i feel that this change dont change anything - you can build with or without this change without issues.

if you can revert this - i will merge this

enthusiastmartin: no longer an issue.

Issue #454, opened 2022-05-18
Change fee payment token on account create, transfer or xcm transfer


jak-pan:
**Description**
If we are creating new account on the chain the fee payment asset could be set to the asset that was transferred to that account. 

**Why**
The idea is that you don't need to do extra transaction when you transfer tokens to new account (disregarding bsx). This would greatly improve UX inside the chain because user doesn't need to do 2 transactions himself (If the account is new, including no system account (or has no fee payment balance) user would need to do second TX. I'm proposing to create a hook that will do this automagically after any transfer / lock. This can be tricky however.

**Possible problems**
- [ ] If the account is programatic (i.e. pool) we might not want to do this.
- [ ] If the account has been created before we also might not want to do this as it can cause confusion that different actions do or not do this for you.
- [x] If somebody spams you with some useless asset it will be changed for you and it will be very confusing.
- [ ] ???

jak-pan:
I would like to get some reasoning here why yes why not. My biggest concern is with spam. It is not as big of a problem since we're always checking your fee payment asset pre-tx signing and prompt you to change it if you don't have any / enough balance.

apopiak:
Based on discussion elsewhere we will want to implement this by adding a callback to ORML tokens (likely [here](https://github.com/open-web3-stack/open-runtime-module-library/blob/39057dae9c277e34a8e1c61abf2148158199e687/tokens/src/lib.rs#L752)) for newly created accounts that includes the asset id and would thus allow us to set it as the fee asset id.

apopiak:
Callback implementation here: https://github.com/open-web3-stack/open-runtime-module-library/pull/754

apopiak:
Callback types here: https://github.com/galacticcouncil/warehouse/pull/54

PR #459, opened 2022-05-24
refactor: events to structs

Roznovjak:
Fixes: #326

PR #476, opened 2022-06-07
feat!: Set Transaction Fee Asset on New Account

apopiak:
Use the types introduced in https://github.com/galacticcouncil/warehouse/pull/54 to set transaction fee asset on new account (and clean up when fee asset balance reaches zero or system account is reaped).

## Description
As described in #454 we want to automatically set a user's fee asset when it makes sense to do so.

We do this by implementing a [new callback introduced to orml_tokens](https://github.com/open-web3-stack/open-runtime-module-library/pull/754) that gets triggered whenever a new account is created for a `(asset_id, account_id)` combination. If the account id does not have a fee currency set and the asset id of the new account is an accepted fee currency, we set the user's fee currency to this new asset.

## Notes
The weights should be rerun because this changes the costs of creating and deleting accounts.

## Related Issue
Fixes: #454

## How Has This Been Tested?

- [x] local integration test
- [x] xcm transfer integration test

## Checklist:


pallets/offchain-duster/src/mock.rs
```diff
@@ -159,6 +159,8 @@ impl orml_tokens::Config for Test {
 	type ExistentialDeposits = ExistentialDeposits;
 	type OnDust = ();
 	type MaxLocks = ();
+	type OnNewTokenAccount = ();
+    type OnKilledTokenAccount = ();
```
apopiak, line 163:
```suggestion
	type OnNewTokenAccount = ();
	type OnKilledTokenAccount = ();
```

Cargo.toml
```diff
@@ -198,8 +198,20 @@ cumulus-relay-chain-interface = { git = "https://github.com/paritytech//cumulus"
...
-pallet-transaction-multi-payment = { git = "https://github.com/galacticcouncil//warehouse", rev = "e428fd81f84781c4508df72f095e7bed6bb58ea1"}
-pallet-asset-registry= { git = "https://github.com/galacticcouncil//warehouse", rev = "e428fd81f84781c4508df72f095e7bed6bb58ea1"}
-hydradx-traits = { git = "https://github.com/galacticcouncil//warehouse", rev = "e428fd81f84781c4508df72f095e7bed6bb58ea1"}
+pallet-price-oracle = { git = "https://github.com/galacticcouncil//warehouse", branch = "apopiak/set-tx-fee-asset-on-new-account"}
+pallet-relaychain-info = { git = "https://github.com/galacticcouncil//warehouse", branch = "apopiak/set-tx-fee-asset-on-new-account"}
+pallet-transaction-multi-payment = { git = "https://github.com/galacticcouncil//warehouse", branch = "apopiak/set-tx-fee-asset-on-new-account"}
+pallet-asset-registry= { git = "https://github.com/galacticcouncil//warehouse", branch = "apopiak/set-tx-fee-asset-on-new-account"}
+hydradx-traits = { git = "https://github.com/galacticcouncil//warehouse", branch = "apopiak/set-tx-fee-asset-on-new-account"}
```
enthusiastmartin, line 205: I would prefer to use a specific revision. So when it gets merged in warehouse master - no change will be needed here.

pallets/xyk/src/tests.rs
```diff
@@ -963,7 +963,12 @@ fn discount_sell_fees_should_work() {
 
 		expect_events(vec![
 			Event::PoolCreated(user_1, asset_a, HDX, 10_000, share_token_native, native_pair_account).into(),
-			pallet_asset_registry::Event::Registered(1, bounded_name, AssetType::PoolShare(asset_a, asset_b)).into(),
+			pallet_asset_registry::Event::Registered {
+				asset_id: 1,
+				asset_name: bounded_name,
+				asset_type: AssetType::PoolShare(asset_a, asset_b),
+			}
+			.into(),
```
enthusiastmartin, line 971: Not part of this PR - this is due to recent changes in warehouse and it was yet updated in Basilisk.

UI needs to be aware of this change.

I wonder it would be better to do it as part of separate PR so it can be easily tracked.

enthusiastmartin (approved):
This looks ok.

I wonder where the update of event structs should be as part of standalone PR  so it can be easily tracked ( because UI needs to be aware ).

pallets/xyk/src/tests.rs
enthusiastmartin: It should be done as part of #459

runtime/common/src/weights/tokens.rs
```diff
@@ -50,28 +50,28 @@ pub struct BasiliskWeight<T>(PhantomData<T>);
 
 impl<T: frame_system::Config> WeightInfo for BasiliskWeight<T> {
 	fn transfer() -> Weight {
-		(66_923_000 as Weight)
-			.saturating_add(T::DbWeight::get().reads(4 as Weight))
-			.saturating_add(T::DbWeight::get().writes(3 as Weight))
+		(66_005_000 as Weight)
```
jak-pan, line 53: Are these extra reads and writes part of our change or is this unrelated orml thing?

enthusiastmartin: i believe it is the extra read and write of "set fee asset" handler.

Cargo.toml
apopiak: Right, this was for development :+1:

runtime/common/src/weights/tokens.rs
apopiak: Jep, that's why I asked to run the weights, cause I wanted us to see the impact.

pallets/xyk/src/tests.rs
apopiak: I guess we'll wait for #459 before merging this?

enthusiastmartin: yes. it will be merged shortly.

lumir-mrkva:
Task linked: [CU-2kpe21t Set Fee Asset Id on New Account](https://app.clickup.com/t/2kpe21t)

apopiak:
requires https://github.com/galacticcouncil/Basilisk-node/pull/493 (and includes its changes)

Issue #486, opened 2022-06-11
Bug: investigate why builds with --all-features fail


enthusiastmartin:
One of the recent changes caused builds with `--all-features` flag to fail ( see output below).
eg.

```
cargo clippy --release --locked --all-targets --all-features
```

Makefile was changed to use `--features=runtime-benchmarks` only whcih works fine but it would be good to figure out what causes this problem and fix if possible.

```
     Compiling sp-io v5.0.0 (https://github.com/paritytech//substrate?rev=22d40c761a985482f93bbbea5ba4199bdba74f8e#22d40c76)
  error[E0152]: found duplicate lang item `panic_impl`
      --> /Users/martin/.cargo/git/checkouts/substrate-f4423eebfa0046a3/22d40c7/primitives/io/src/lib.rs:1629:1
       |
  1629 | pub fn panic(info: &core::panic::PanicInfo) -> ! {
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       |
       = note: the lang item is first defined in crate `std` (which `tracing_core` depends on)
       = note: first definition in `std` loaded from /Users/martin/.rustup/toolchains/nightly-2022-02-19-x86_64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libstd-d9a8e4ca4a56d589.rlib
       = note: second definition in the local crate (`sp_io`)

  error[E0152]: found duplicate lang item `oom`
      --> /Users/martin/.cargo/git/checkouts/substrate-f4423eebfa0046a3/22d40c7/primitives/io/src/lib.rs:1645:1
       |
  1645 | pub fn oom(_: core::alloc::Layout) -> ! {
       | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       |
       = note: the lang item is first defined in crate `std` (which `tracing_core` depends on)
       = note: first definition in `std` loaded from /Users/martin/.rustup/toolchains/nightly-2022-02-19-x86_64-apple-darwin/lib/rustlib/wasm32-unknown-unknown/lib/libstd-d9a8e4ca4a56d589.rlib
       = note: second definition in the local crate (`sp_io`)

  For more information about this error, try `rustc --explain E0152`
```

enthusiastmartin:
Resolved in #488 .

We can put back the "--all-features" flag in Makefile.

PR #487, opened 2022-06-11
ci: fix docs worklow - build only with runtime-benchmarks features

enthusiastmartin:
Related to #486 

This might be a temporary fix while the issue is being investigated.


enthusiastmartin:
Closing in favor of #488

PR #488, opened 2022-06-13
fix: docs build

Roznovjak:


enthusiastmartin (approved):
LGTM if passes.

You might then put back the "--all-features" flag in Makefile (as part of this #486 )

Roznovjak:
> LGTM if passes.
> 
> You might then put back the "--all-features" flag in Makefile (as part of this #486 )

Do we want to do it as part of this PR?

enthusiastmartin:
> > LGTM if passes.
> > You might then put back the "--all-features" flag in Makefile (as part of this #486 )
> 
> Do we want to do it as part of this PR?

No. in a new one.

Issue #412, opened 2022-03-23
Fee's for execution goes to Treasury


lumir-mrkva:

jak-pan:
blocked by https://github.com/galacticcouncil/warehouse/pull/42

PR #493, opened 2022-06-14
feat!: Transfer Fees to Treasury

apopiak:
Use the `TransferFees` type introduced in https://github.com/galacticcouncil/warehouse/pull/42 in the runtimes and fix various type/integration issues.

## Related Issue

Fixes: #412

## Motivation and Context

This transfers fees to the treasury instead of burning them.
In addition it also unblocks other PRs depending on a recent version of the warehouse.

## Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -198,8 +198,8 @@ cumulus-relay-chain-interface = { git = "https://github.com/paritytech//cumulus"
 cumulus-relay-chain-local = { git = "https://github.com/paritytech//cumulus", rev = "76479e7fef3af7c8828a44647847b01afd5fefe5" }
 
 [patch."https://github.com/galacticcouncil/warehouse"]
-pallet-price-oracle = { git = "https://github.com/galacticcouncil//warehouse", rev = "50c82ab02bb5baf99b263342dc5d4cc7e62c7e88"}
-pallet-relaychain-info = { git = "https://github.com/galacticcouncil//warehouse", rev = "50c82ab02bb5baf99b263342dc5d4cc7e62c7e88"}
-pallet-transaction-multi-payment = { git = "https://github.com/galacticcouncil//warehouse", rev = "50c82ab02bb5baf99b263342dc5d4cc7e62c7e88"}
-pallet-asset-registry= { git = "https://github.com/galacticcouncil//warehouse", rev = "50c82ab02bb5baf99b263342dc5d4cc7e62c7e88"}
-hydradx-traits = { git = "https://github.com/galacticcouncil//warehouse", rev = "50c82ab02bb5baf99b263342dc5d4cc7e62c7e88"}
+pallet-price-oracle = { git = "https://github.com/galacticcouncil//warehouse", rev = "7052eeab309742dccfe7e307c15fc4f780713553"}
+pallet-relaychain-info = { git = "https://github.com/galacticcouncil//warehouse", rev = "7052eeab309742dccfe7e307c15fc4f780713553"}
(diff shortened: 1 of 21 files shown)


runtime/testing-basilisk/src/lib.rs
```diff
@@ -304,6 +308,7 @@ impl pallet_transaction_multi_payment::Config for Runtime {
 	type WithdrawFeeForSetCurrency = MultiPaymentCurrencySetFee;
 	type WeightToFee = WeightToFee;
 	type NativeAssetId = ();
```
enthusiastmartin, line 310: hm.not part of this PR: but why there is a missing native asset id ? i wonder how this could work without it 🤔

enthusiastmartin (commented):
it looks good to me. it is just an integration of work which hopefully has been done and tested. 

I'll leave to @Roznovjak to approve.

runtime/testing-basilisk/src/lib.rs
Roznovjak: https://github.com/paritytech/substrate/blob/4037345b8c0491ae5b74fcb6d131b3ff304f0bed/primitives/runtime/src/traits.rs#L297

Roznovjak (approved):
Looks good. There is probably no documentation requiring changes. But I'm not sure about tests. Do we want to add some tests? The problem with testing is that I'm not sure how we can call a TX and charge a fee for it without using dedicated struct for handling fees (`ChargeTransactionPayment`, as used in the tests in the TX multi payment pallet).

runtime/testing-basilisk/src/lib.rs
enthusiastmartin: well.that means we just lucky that native is 0 then. 

better to fix it.

enthusiastmartin: #495

Cargo.toml
```diff
@@ -198,8 +198,8 @@ cumulus-relay-chain-interface = { git = "https://github.com/paritytech//cumulus"
 cumulus-relay-chain-local = { git = "https://github.com/paritytech//cumulus", rev = "76479e7fef3af7c8828a44647847b01afd5fefe5" }
 
 [patch."https://github.com/galacticcouncil/warehouse"]
-pallet-price-oracle = { git = "https://github.com/galacticcouncil//warehouse", rev = "c73846ec093357fea1b35da5c56ecb37592d4896"}
-pallet-relaychain-info = { git = "https://github.com/galacticcouncil//warehouse", rev = "c73846ec093357fea1b35da5c56ecb37592d4896"}
-pallet-transaction-multi-payment = { git = "https://github.com/galacticcouncil//warehouse", rev = "c73846ec093357fea1b35da5c56ecb37592d4896"}
-pallet-asset-registry= { git = "https://github.com/galacticcouncil//warehouse", rev = "c73846ec093357fea1b35da5c56ecb37592d4896"}
-hydradx-traits = { git = "https://github.com/galacticcouncil//warehouse", rev = "c73846ec093357fea1b35da5c56ecb37592d4896"}
+pallet-price-oracle = { git = "https://github.com/galacticcouncil//warehouse", rev = "50c82ab02bb5baf99b263342dc5d4cc7e62c7e88"}
```
enthusiastmartin, line 201: Why we dont use the very latest warehouse revision? 

this one : 7052eeab309742dccfe7e307c15fc4f780713553

enthusiastmartin: I have updated to the latest revision.

PR #497, opened 2022-06-16
chore: use pallet-nft from warehouse

enthusiastmartin:
Repleace local nft pallet with the one from warehouse repository.


pallets/marketplace/src/tests.rs
```diff
@@ -293,12 +292,15 @@ fn offering_works() {
 			INSTANCE_ID_0,
 			metadata.clone()
 		));
+		/*
```
green-jay, line 295: there is nothing done with class 1 in the test so we can remove this assertion

green-jay (approved):

pallets/marketplace/src/tests.rs
enthusiastmartin: done

PR #499, opened 2022-06-24
fix: xyk shares calculation

enthusiastmartin:
Shares calculation in add_liquidity has been changed. 

it is now calculated as follows:

delta_shares = share_issuance * delta_X / reserve_x 

where X is an asset provided by LP. 

it was also moved to the math crate so xyk now calls only calculate_shares from the xyk math module. 

requires https://github.com/galacticcouncil/HydraDX-math/pull/42/files

### DIFF
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -421,13 +421,8 @@ pub mod pallet {
 
 			ensure!(amount_b <= amount_b_max_limit, Error::<T>::AssetAmountExceededLimit);
 
-			let shares_added = if asset_a < asset_b {
-				hydra_dx_math::xyk::calculate_shares(asset_a_reserve, amount_a, share_issuance)
-					.ok_or(Error::<T>::Overflow)?
-			} else {
-				hydra_dx_math::xyk::calculate_shares(asset_b_reserve, amount_b, share_issuance)
-					.ok_or(Error::<T>::Overflow)?
-			};
+			let shares_added = hydra_dx_math::xyk::calculate_shares(asset_a_reserve, amount_a, share_issuance)
+				.ok_or(Error::<T>::Overflow)?;
 
 			ensure!(!shares_added.is_zero(), Error::<T>::InvalidMintedLiquidity);
 
--- a/pallets/xyk/src/lib.rs
+++ b/pallets/xyk/src/lib.rs
@@ -414,18 +414,20 @@ pub mod pallet {
 
 			let asset_a_reserve = T::Currency::free_balance(asset_a, &pair_account);
 			let asset_b_reserve = T::Currency::free_balance(asset_b, &pair_account);
-			let total_liquidity = Self::total_liquidity(&pair_account);
+			let share_issuance = Self::total_liquidity(&pair_account);
 
-			let amount_b_required =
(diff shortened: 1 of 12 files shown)


pallets/xyk/src/tests.rs
```diff
@@ -872,9 +873,17 @@ fn work_flow_happy_path_should_work() {
...
 				asset_b,
 				amount_a: 300_000_000_000,
-				amount_b: 12_000_000_000_000,
+				amount_b: 12_000_000_000_001,
 			}
 			.into(),
+			/*
+
+			// TODO:
+			// Note: removing these events because it is very hard to determine the correct values
```
lumir-mrkva, line 882: i would just remove it

pallets/xyk/src/tests.rs
```diff
@@ -2200,3 +2211,103 @@ fn get_fee_should_work() {
 		assert_eq!(fee, (2, 1_000));
 	});
 }
+
+#[test]
+fn bug_scenario() {
```
lumir-mrkva, line 2216: cannot_drain_pool()

lumir-mrkva (approved):

pallets/xyk/src/lib.rs
```diff
@@ -414,18 +414,20 @@ pub mod pallet {
...
+				.map_err(|_| Error::<T>::AddAssetAmountInvalid)?;
 
-			let shares_added = if asset_a < asset_b { amount_a } else { amount_b_required };
+			ensure!(amount_b <= amount_b_max_limit, Error::<T>::AssetAmountExceededLimit);
 
-			ensure!(
-				amount_b_required <= amount_b_max_limit,
-				Error::<T>::AssetAmountExceededLimit
-			);
+			let shares_added = if asset_a < asset_b {
```
poliwop, line 424: this if clause is not necessary, calculate_shares(asset_a_reserve, amount_a, share_issuance) and calculate_shares(asset_b_reserve, amount_b, share_issuance) should return the same result (up to rounding error).

Probably best to always use asset a, since amount_b is rounded up and compounding rounding errors could give user too many shares.

Somewhere there should perhaps be a property test that the calculation with assets a & b do in fact return the same result, not sure what exactly the right place for that would be.

pallets/xyk/src/tests.rs
```diff
@@ -2200,3 +2158,103 @@ fn get_fee_should_work() {
 		assert_eq!(fee, (2, 1_000));
 	});
 }
+
+#[test]
+fn cannot_drain_pool() {
```
poliwop, line 2163: this test is actually a very, very specific case... I'd suggest something more like "fn share_ratio_calculations_are_correct" even though that's still pretty generic, it's more targetted. This test is definitely not sufficient to guarantee that the pool cannot be drained.

poliwop (changes requested):

pallets/xyk/src/lib.rs
enthusiastmartin: thanks. changed and tested. Also added one simple test with exact values to make sure that it returns same shares.

i will add a property test for that  too in #501

poliwop (approved):
Looks good to me.

PR #500, opened 2022-06-24
refactor: discounted fee

Roznovjak:
Make the discounted fee configurable instead of using hardcoded value.

### DIFF
--- a/pallets/xyk/Cargo.toml
+++ b/pallets/xyk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-xyk'
-version = '3.7.0'
+version = '4.0.0'
 description = 'XYK automated market maker'
 authors = ['GalacticCouncil']
 edition = '2021'
--- a/pallets/xyk/Cargo.toml
+++ b/pallets/xyk/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-xyk'
-version = '3.6.4'
+version = '3.7.0'
 description = 'XYK automated market maker'
 authors = ['GalacticCouncil']
 edition = '2021'
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -486,7 +486,7 @@ dependencies = [
 
 [[package]]
 name = "basilisk-runtime"
-version = "58.0.0"
+version = "59.0.0"
 dependencies = [
  "common-runtime",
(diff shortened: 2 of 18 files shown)


lumir-mrkva:
Task linked: [CU-2jfyk51 XYK - refactor hardcoded discounted fee](https://app.clickup.com/t/2jfyk51)

enthusiastmartin (changes requested):
I'd suggest to add few more tests with different fees (zero included).

Now it is set to the original default value 0.07% but it is not really tested whether it works with different values.

Roznovjak:
> I'd suggest to add few more tests with different fees (zero included).
> 
> Now it is set to the original default value 0.07% but it is not really tested whether it works with different values.

We don't test config types for various values because it would require a lot of mocked runtimes.
But the `DiscountedFee` is used only in one method, and this method is properly [tested](https://github.com/galacticcouncil/HydraDX-math/blob/33c7749147fa5766f4e1cbf2019bfa654d3683aa/src/fee.rs#L90).

enthusiastmartin:
> > I'd suggest to add few more tests with different fees (zero included).
> > Now it is set to the original default value 0.07% but it is not really tested whether it works with different values.
> 
> We don't test config types for various values because it would require a lot of mocked runtimes. But the `DiscountedFee` is used only in one method, and this method is properly [tested](https://github.com/galacticcouncil/HydraDX-math/blob/33c7749147fa5766f4e1cbf2019bfa654d3683aa/src/fee.rs#L90).

you can test similar way how exchange fee is tested for xyk pallet. 

you can easily build tested runtime with different fee.

pallets/xyk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-xyk'
-version = '3.6.5'
+version = '3.7.0'
```
enthusiastmartin, line 3: i think this should be a major version change . 

Because it introduces a change which is non backward compatible.wdyt ?

Roznovjak: Not sure if a pallet Config belongs to public API. But we can bump the major version.

enthusiastmartin: sure it does. you can upgrade pallet-xyk version in runtime as the config changes.

PR #501, opened 2022-06-24
feat: xyk invariant tests

enthusiastmartin:
This PR introduces new XYK property tests for add/remove liquidity where certain properties must hold after LP adds or removes liquidity. 

It also introduces invariant tests for XYK trades where asset invariant must be after sell or buy.

Tests are reorganized as part of this PR where all test related stuff is now location in tests sub directory. This is mainly done for better readability and to easily understand what is part of tests and what is pallet's source. 

- [x] add test to verify that add liquidity mints corrects same shares when LP provides asset a or asset b.


pallets/xyk/src/invariants.rs
```diff
@@ -0,0 +1,445 @@
...
+				let delta_s = U256::from(bob_shares);
+				let delta_x = U256::from(bob_balance_a - new_bob_balance_a);
+				let delta_y = U256::from(bob_balance_b - new_bob_balance_b);
+				let x = U256::from(pool_balance_a);
+				let y = U256::from(pool_balance_b);
+
+				let l = delta_s * x;
+				let r = s * delta_x;
+
+				assert!(l <= r);
```
poliwop, line 129: It may be good to also add a check here that assert_eq_approx!(l, r, ...), for both the check with x and check with y.

These inequality checks assert that any discrepancy betweten l and r is in favor of the pool, but we also want l and r to be equal (up to rounding errors).

pallets/xyk/src/invariants.rs
```diff
@@ -0,0 +1,445 @@
...
+				let delta_s = U256::from(bob_shares);
+				let delta_x = U256::from(new_bob_balance_a - bob_balance_a);
+				let delta_y = U256::from(new_bob_balance_b - bob_balance_b);
+				let x = U256::from(pool_balance_a);
+				let y = U256::from(pool_balance_b);
+
+				let l = delta_s * x;
+				let r = s * delta_x;
+
+				assert!(l >= r);
```
poliwop, line 223: I would suggest testing assert_eq_approx(l, r, ...) for both x and y

pallets/xyk/src/invariants.rs
```diff
@@ -0,0 +1,445 @@
...
+						asset_a,
+						amount,
+						u128::MAX, // limit not interesting here,
+						true,
+				));
+
+				let _new_pool_balance_a = Currency::free_balance(asset_a, &pool_account);
+				let _new_pool_balance_b = Currency::free_balance(asset_b, &pool_account);
+
+				//TODO: this currently fails - under investigation
```
poliwop, line 433: discounted fee is hardcoded at 0.07%, this invariant only holds with fees set to zero.

poliwop (commented):
Suggested a couple more tests, otherwise looks good.

pallets/xyk/src/invariants.rs
enthusiastmartin: Done

pallets/xyk/src/invariants.rs
enthusiastmartin: Need #500 , then we can set fee to 0 and test.

pallets/xyk/src/tests/tests.rs
```diff
@@ -15,11 +15,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::*;
-pub use crate::mock::{
+pub use super::mock::{
 	Currency, Event as TestEvent, ExtBuilder, Origin, System, Test, ACA, ALICE, BOB, CHARLIE, DOT, HDX,
 	HDX_DOT_POOL_ID, ONE, XYK,
 };
+use crate::*;
```
lumir-mrkva, line 22: this file should be renamed `lib.rs`

lumir-mrkva (changes requested):

pallets/xyk/src/tests/tests.rs
enthusiastmartin: done

pallets/xyk/src/invariants.rs
Roznovjak: #500 should be ready. At least for review.

Roznovjak:
Just a note: There is only `buy_invariant_with_discount` and no test for sell. If you are aware of it or if it's ok, please ignore this comment.

enthusiastmartin:
> Just a note: There is only `buy_invariant_with_discount` and no test for sell. If you are aware of it or if it's ok, please ignore this comment.

added

Roznovjak (approved):

PR #503, opened 2022-06-30
fix: use native asset in tx multi payment config

### LINKED ISSUE #495: Fix invalid native asset id in testing-runtime (tx-payment pallet)
NativeAssetId is not configured for tx multi payment pallet  in testing runtime.

However it works ok as it takes default value and it happens that native is the default value. 

it would be better to set it correctly to avoid any surprises.

hm.not part of this PR: but why there is a missing native asset id ? i wonder how this could work without it 🤔

_Originally posted by @enthusiastmartin in https://github.com/galacticcouncil/Basilisk-node/pull/493#discussion_r897281447_

Roznovjak:
Fixes #495


Roznovjak:
I'm not sure if we want to update runtime versions for such a small update in testing runtime

enthusiastmartin (approved):

PR #512, opened 2022-07-08
feat: NFT marketplace uses relay chain currency

lumir-mrkva:
also little bit of test changes that we did on Friday hack

dependent of https://github.com/galacticcouncil/warehouse/pull/65

### DIFF
--- a/runtime/basilisk/src/lib.rs
+++ b/runtime/basilisk/src/lib.rs
@@ -50,6 +50,9 @@ use sp_std::prelude::*;
 use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 
+#[cfg(feature = "runtime-benchmarks")]
+use codec::{Decode, Encode};
+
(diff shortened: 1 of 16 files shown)


runtime/basilisk/src/lib.rs
```diff
@@ -444,17 +445,6 @@ impl pallet_duster::Config for Runtime {
 	type WeightInfo = common_runtime::weights::duster::BasiliskWeight<Runtime>;
 }
 
-/// Basilisk Pallets configurations
-
-#[derive(Debug, Encode, Decode, Clone, PartialEq, Eq, TypeInfo)]
```
jak-pan, line 449: Why is this removed?

jak-pan:
LGTM I didn't look at side effects that could happen but believe if tests pass it should be gtg

jak-pan (approved):

PR #517, opened 2022-07-11
feat: GC distribution

lumir-mrkva:


scripts/distribute-into-proxies/index.js
```diff
@@ -0,0 +1,201 @@
...
+  assert.equal(
+    anonymousProxies.length,
+    distribution.length,
+    'not all proxies created',
+  )
+  log('proxies created:', anonymousProxies)
+
+  log('funding proxies...')
+  const transfers = anonymousProxies.map((anon) =>
+    api.tx.balances.transfer(anon, 500 * UNIT),
```
jak-pan, line 132: Is this enough after fee bump? We probably also need sudo here if we do it b4 upgrade.

scripts/distribute-into-proxies/index.js
```diff
@@ -0,0 +1,201 @@
...
+
+let proxyIndex = 2000
+const multisig = 'bXiGFV6H2JpQYqXFtQtRqXNn7zS22Nco349BE7CvUzBuidraU'
+const UNIT = 1000000000000
+
+const vesting = {
+  start: '13517962',
+  period: '1752',
+  per_period: '',
+  period_count: '6000',
```
jak-pan, line 20: Changed so there are no remainders

scripts/distribute-into-proxies/index.js
lumir-mrkva: its for storage deposit, which shouldn't change, but we'll test

lumir-mrkva:
prettier sucks, stealing my perfect code...

PR #523, opened 2022-07-13
test: fees integration tests fixed

martinfridrich:
Added checks for fees from polkadotAPPS UI and missing tests.


integration-tests/src/fees.rs
```diff
@@ -3,35 +3,85 @@
 use crate::kusama_test_net::*;
 use basilisk_runtime::TransactionPayment;
 use frame_support::weights::GetDispatchInfo;
+use primitives::nft::ClassType;
 use sp_runtime::codec::Encode;
 use xcm_emulator::TestExt;
 
+macro_rules! assert_eq_approx {
```
dmoka, line 10: This macro is duplicated as it can be also found in `xy/src/tests/invariants.rs`

I wonder if we could start extracting such test helper logics into some common test-util crate, so we could remove duplications...What do you think?

integration-tests/src/fees.rs
```diff
@@ -3,35 +3,85 @@
...
+		}
+	}};
+}
+
+//NOTE: rust encoded call size is differen from UI encoded call size that's why we have asserts for 2 fees.
+
 #[test]
-// #[ignore]
-//TODO: fix this, fees calculcated in this test are apporx 2bsx off.
 fn transfer_transaction_fees() {
```
dmoka, line 22: I would improve the test name according to our ShouldWhen test naming template. ;)

Also applies for the other tests in this file.

integration-tests/src/fees.rs
```diff
@@ -3,35 +3,85 @@
...
-	//This test is not correct
 	Basilisk::execute_with(|| {
+		let diff = UNITS / 100; //0.01
+
+		let expected_rust_encoded_fees = 45_56 * UNITS / 100; //45.56
+		let expected_ui_fees = 4_655 * UNITS / 100; //46.55
```
dmoka, line 27: what is the reason for having these comments here, describing the values of the variables?

The thing is that once the UNITS changes, then these comment can easily become outdated.

integration-tests/src/fees.rs
```diff
@@ -3,35 +3,85 @@
...
+
+		assert_eq_approx!(
+			rust_encoded_fees,
+			expected_rust_encoded_fees,
+			diff,
+			"Rust encoded fees are"
+		);
+
+		//UI encoded fees
+		let len = 143; //UI encoded length
```
dmoka, line 48: I would call this variable as `ui_encoded_length`, then the comment is not necessary.

integration-tests/src/fees.rs
```diff
@@ -3,35 +3,85 @@
...
+		pallet_transaction_payment::pallet::NextFeeMultiplier::<basilisk_runtime::Runtime>::put(
+			basilisk_runtime::MinimumMultiplier::get(),
+		);
 		let len = call.encoded_size() as u32;
-		let fees = TransactionPayment::compute_fee(len, &info, 0);
+		let min_multiplier_rust_fees = TransactionPayment::compute_fee(len, &info, 0);
 
-		//This test is not correct it's approx 1bsx of from real fees
-		println!("transfer cost {:?} BSX", fees / UNITS);
+		let len = 143; //UI encoded length
```
dmoka, line 59: we already have this variable present earlier, so I think it is not necessary to re-declare here.

integration-tests/src/fees.rs
```diff
@@ -3,35 +3,85 @@
...
+		let min_multiplier_ui_fees = TransactionPayment::compute_fee(len, &info, 0);
+
+		println!(
```
dmoka, line 62: why do we print all the fee values at the end of these tests? Do we need verify it visually?

dmoka (dismissed):

integration-tests/src/fees.rs
martinfridrich: yes, It describe expected fees value in readable form. `UNITS` can't change without this test failing so whoever will be fixing this test can update comments.

integration-tests/src/fees.rs
dmoka: and what is the reason for that? 

I mean, why it is better than using some assertions in the code?

martinfridrich: There are assertions in the code but they are not precise. It's up to us said if the diff is ok. 
Also when you are changing fees it's easier to see them, than running whole chain and look into UI.
This things are not printed by default so I don't see any downside of having them

dmoka: @martinfridrich fair enough, sound reasonable then!

integration-tests/Cargo.toml
```diff
@@ -108,6 +108,7 @@ kusama-runtime = { git = "https://github.com/paritytech/polkadot", branch = "rel
 
 [dev-dependencies]
 xcm-emulator = { git = "https://github.com/zqhxuyuan/xcm-simulator", branch = "v0.9.16" }
+sp-arithmetic = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
```
apopiak, line 111: nitpick: missing in the std feature list, or alternatively include without `default-features = false` as it's a dev-dependency

integration-tests/src/fees.rs
```diff
@@ -3,35 +3,82 @@
...
+			rust_encoded_fees,
+			expected_rust_encoded_fees,
+			diff,
+			"Rust encoded fees are"
+		);
 
-		//This test is not correct it's approx 1bsx of from real fees
-		println!("transfer cost {:?} BSX", fees / UNITS);
+		//UI encoded fees
+		let ui_encoded_len = 143;
```
apopiak, line 48: any way we can notice if this should change?
maybe add an `assert!(ui_encoded_len > rust_encoded_len)` as a sanity check?

integration-tests/src/fees.rs
```diff
@@ -40,10 +87,103 @@ fn nft_mint_transaction_fees() {
...
+
+		println!(
+			"NFT create_class\n\t UI fees: {}/{} [actual/expected]\n\t Rust encoded fees: {}/{} [actual/expected]\n\t Fees with min. FeeMultiplier: {} [UI], {} [Rust]",
+			format_num(ui_fees * 10_000 / UNITS, 4),
+			format_num(expected_ui_fees * 10_000 / UNITS, 4),
+			format_num(rust_encoded_fees * 10_000 / UNITS, 4),
+			format_num(expected_rust_encoded_fees * 10_000 / UNITS, 4),
+			format_num(min_multiplier_ui_fees * 10_000 / UNITS, 4),
+			format_num(min_multiplier_rust_fees * 10_000 / UNITS, 4),
+		);
```
apopiak, line 179: dumb question: will this ever be printed?
if the asserts fail, this code won't execute, right?
and if no assert fails, then output is not printed

do you expect to execute with `--no-capture` when interested in the print-out?

integration-tests/src/fees.rs
dmoka: I wonder what would be the best place for that. I am not sure that placing it in Basilisk would be the best as the helpers might be used in other projects. Maybe warehouse repo, or even better to have a dedicated repo for this. What do you think?

integration-tests/src/fees.rs
martinfridrich: yes, you are right. 
I expect to use `--no-capture` if you are interested to see values with more precision or if you are setting "new fees" you can temporally comment asserts and look into values(you don't have to check it in the UI) and update expected values last.

martinfridrich: but true, asserts can be done after printing and commenting will not be necessary, I'll do it

integration-tests/src/fees.rs
```diff
@@ -40,10 +88,103 @@ fn nft_mint_transaction_fees() {
...
 		let info = call.get_dispatch_info();
-		let len = call.encoded_size() as u32;
-		let fees = TransactionPayment::compute_fee(len, &info, 0);
+		//rust encoded fees
+		let rust_encoded_len = call.encoded_size() as u32;
+		let rust_encoded_fees = TransactionPayment::compute_fee(rust_encoded_len, &info, 0);
+
+		//UI encoded fees
+		let ui_encoded_len = 192;
+		assert_eq!(ui_encoded_len - rust_encoded_len, 105);
```
apopiak, line 97: let's also add an assert message or comment?
something like "lengths expected to differ by a constant value at test creation, revisit if encoded call length changes")

apopiak (approved):
There's some inherent ugliness in a bunch of the hard-coding for the tests.
I think you made the best of a bad situation :-D

integration-tests/src/fees.rs
lumir-mrkva: lets create warehouse crate for these testing utils, i am sure we can find more stuff that could be extracted there

lumir-mrkva: pls. create issue for that :)

integration-tests/src/fees.rs
```diff
@@ -3,35 +3,83 @@
...
+			format_num(expected_rust_encoded_fees * 10_000 / UNITS, 4),
+			format_num(min_multiplier_ui_fees * 10_000 / UNITS, 4),
+			format_num(min_multiplier_rust_fees * 10_000 / UNITS, 4),
+		);
+
+		assert_eq_approx!(
+			rust_encoded_fees,
+			expected_rust_encoded_fees,
+			diff,
+			"Rust encoded fees are"
```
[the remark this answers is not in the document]
lumir-mrkva: changed

integration-tests/src/fees.rs
lumir-mrkva: adjusted

lumir-mrkva (approved):

PR #527, opened 2022-07-14
feat: add XYK liquidity mining pallet

dmoka:
## Description

Added implementation for XYK liquidity mining where we use the extracted liquidity mining pallet from warehouse repo.

TODO: 
- it is still not possible to create nft class for LM, so the migration/benchmark tests are not final
- because of the same reason, the weights are not generated yet

## Checklist:


pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	fn account_id_for_all_lp_shares() -> AccountIdOf<T> {
+		<T as pallet::Config>::PalletId::get().into_account()
+	}
+
+	fn get_lp_token(amm_pool_id: &AccountIdOf<T>) -> Result<AssetId, Error<T>> {
```
martinfridrich, line 868: I don't think this fn is necessary. You have `AssetPair` as part of api calls, conver it to `amm_pool_id` and back is unnecessary.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+				});
+			}
+
+			Ok(())
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	fn account_id_for_all_lp_shares() -> AccountIdOf<T> {
```
martinfridrich, line 864: I don't like this name. This account may be used for multiple things in the future.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+		let lp_token = Self::get_lp_token(&amm_pool_id)?;
+
+		let service_account_for_lp_shares = Self::account_id_for_all_lp_shares();
+		MultiCurrencyOf::<T>::transfer(lp_token, &who, &service_account_for_lp_shares, amount)?;
+
+		Ok(())
+	}
+
+	fn unlock_lp_tokens(amm_pool_id: T::AccountId, who: T::AccountId, amount: Balance) -> Result<(), DispatchError> {
+		let lp_token = Self::get_lp_token(&amm_pool_id)?;
```
martinfridrich, line 890: same as before, this fn is not necessary in my opinion.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+
+		let service_account_for_lp_shares = Self::account_id_for_all_lp_shares();
+		MultiCurrencyOf::<T>::transfer(lp_token, &service_account_for_lp_shares, &who, amount)?;
+
+		Ok(())
+	}
+
+	fn mint_nft_representing_deposit(who: T::AccountId, deposit_id: &u128) -> Result<(), DispatchError> {
```
martinfridrich, line 898: as mentioned, use nft call direclty and remove this fn.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+		T::NFTHandler::mint_into(&T::NftClassId::get(), deposit_id, &who)?;
+
+		Ok(())
+	}
+
+	fn get_asset_balance_in_liquidity_mining_pool(
```
martinfridrich, line 904:
```suggestion
	fn get_asset_balance_in_amm_pool(
```
This fn should return amount of incentivized asset in amm pool.

martinfridrich (commented):
I went only through `pallets/liquidity-mining/src/lib.rs`. This  is not full review. More doc comments probably needs update and it require second look.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+		/// liq. mining program have to have `incentivized_asset` in their pair.
+		/// - `reward_currency`: payoff currency of rewards.
+		/// - `owner`: liq. mining farm owner.
+		/// - `yield_per_period`: percentage return on `reward_currency` of all pools p.a.
+		/// - `min_deposit`: minimum amount which can be deposited to the farm
+		/// - `price_adjustment`:
+		/// Emits `FarmCreated` event when successful.
+		//TODO: Ask Martin what price_adjustment is
+		#[allow(clippy::too_many_arguments)]
+		#[pallet::weight(<T as Config>::WeightInfo::create_farm())]
```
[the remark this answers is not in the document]
dmoka: yes, I did not rename these yet as it does not compile if I do so. I will rename these once the weights are regenerated.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,910 @@
...
+		#[pallet::weight(<T as Config>::WeightInfo::update_liquidity_pool())]
+		#[transactional]
+		pub fn update_yield_farm(
+			origin: OriginFor<T>,
+			farm_id: GlobalFarmId,
+			asset_pair: AssetPair,
+			multiplier: PoolMultiplier,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
```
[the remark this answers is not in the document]
dmoka: Good catch! I will add these checks to other methods as well where asset_pair is used to retrieve the pool id.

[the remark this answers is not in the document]
martinfridrich: Don't add it everywhere. Admin needs ability to stop and destroy `YieldFarm` if amm doesn't exists anymore.

pallets/liquidity-mining/src/lib.rs
dmoka: What I like in this name is that it clearly describes what it is used for, so it captures the domain knowledge. If it will be used for multiple things in the future, we can always refactor/rename/generalize it accordingly.

pallets/liquidity-mining/src/lib.rs
dmoka: good one, it can be indeed simplified.

pallets/liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,893 @@
...
+		///Deposit data not found
+		DepositDataNotFound,
+	}
+
+	#[pallet::event]
+	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
+	pub enum Event<T: Config> {
+		/// New global farm was created.
+		GlobalFarmCreated {
```
martinfridrich, line 182: add: total rewards, min. deposit, price adjustment

pallets/liquidity-mining/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+serde = { features = ["derive"], optional = true, version = "1.0.101" }
+sp-arithmetic = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
+pallet-uniques = { git = "https://github.com/paritytech/substrate", branch = "polkadot-v0.9.17", default-features = false }
+
+
+# ORML dependencies
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", rev = "aac79b3b31953381669a2ffa9b3e9bfe48e87f38", default-features = false }
+
+# HydraDX dependencies
+warehouse_liquidity_mining = { package="pallet-liquidity-mining", git = "https://github.com/galacticcouncil/warehouse", rev="3230d9d1e6eaed6305b1b67587437415cd34e268", default-features = false }
```
apopiak, line 26: nitpick: convention is to use `-` in crate names in toml files
```suggestion
warehouse-liquidity-mining = { package="pallet-liquidity-mining", git = "https://github.com/galacticcouncil/warehouse", rev="3230d9d1e6eaed6305b1b67587437415cd34e268", default-features = false }
```

enthusiastmartin (commented):
i havent looked at tests. 

however, i have one additional concern and that's dusting of account where shares are transferred. 

it should be excluded from dusting imo. eg. xyk pool accounts are excluded.

pallets/xyk-liquidity-mining/Cargo.toml
```diff
@@ -0,0 +1,65 @@
+[package]
+name = "pallet-xyk-liquidity-mining"
+version = "1.3.0"
```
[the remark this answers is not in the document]
martinfridrich: previous version was in the runtime but it was never really used so maybe we can use `1.0.0`

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,938 @@
...
+
+			let mut unclaimable_rewards: Balance = 0;
+			if T::LiquidityMiningHandler::is_yield_farm_claimable(global_farm_id, yield_farm_id, amm_pool_id.clone()) {
+				//This should not fail on double claim, we need unclaimable_rewards
+				let fail_on_double_claim = true;
+				let (global_farm_id, reward_currency, claimed, unclaimable) = T::LiquidityMiningHandler::claim_rewards(
+					who.clone(),
+					deposit_id,
+					yield_farm_id,
+					!fail_on_double_claim,
```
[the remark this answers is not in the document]
martinfridrich: Users can claim rewards without withdrawing LP shares. This should fail if the user called claim 2nd times in the same period. 
Withdrawing shares is claiming automatically and we don't want to penalise users by waiting for the next period to withdraw if the user has already claimed. So in this case 2nd claim should not fail.

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,931 @@
...
+		}
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	/// Account ID of the pot holding locked LP shares. This account is also owner of NFT class
+	/// for all the NFTs minted by this pallet.
+	fn account_id() -> Option<T::AccountId> {
+		<T as pallet::Config>::PalletId::get().try_into_account()
+	}
```
apopiak, line 900: might be fine to use `into_account_truncating` or whatever it's called if you know that `PalletId` always fits into the `AccountId` type.

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,931 @@
...
+
+	fn get_lp_token(amm_pool_id: &T::AccountId) -> Result<AssetId, Error<T>> {
+		let assets = T::AMM::get_pool_assets(amm_pool_id).ok_or(Error::<T>::CantGetAmmAssets)?;
+		let asset_pair = AssetPair::new(assets[0], assets[1]);
+
+		//NOTE: this check is important AMM:get_share_token() return `0` if amm doesn't exist
+		if !T::AMM::exists(asset_pair) {
+			return Err(Error::<T>::AmmPoolDoesNotExist);
+		}
```
apopiak, line 909: I would expect
```suggestion
		ensure!(T::AMM::exists(asset_pair), Error::<T>::AmmPoolDoesNotExist);
```
to work here.
Maybe you would need to adjust the Error in the result above to `DispatchError`, not sure.

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,931 @@
...
+
+		Ok(T::AMM::get_share_token(asset_pair))
+	}
+
+	fn lock_lp_tokens(lp_token: AssetId, who: &T::AccountId, amount: Balance) -> Result<(), DispatchError> {
+		let service_account_for_lp_shares = Self::account_id().ok_or(Error::<T>::ErrorGetAccountId)?;
+		T::MultiCurrency::transfer(lp_token, who, &service_account_for_lp_shares, amount)?;
+
+		Ok(())
```
apopiak, line 918:
```suggestion
		T::MultiCurrency::transfer(lp_token, who, &service_account_for_lp_shares, amount)
```
maybe?

pallets/xyk-liquidity-mining/src/migration.rs
```diff
@@ -0,0 +1,92 @@
...
+				StorageVersion::new(STORAGE_VERSION).put::<Pallet<T>>();
+			}
+		} else {
+			log::error!(
+				target: "basilisk:liquidity-mining",
+				"Error to get pallet account",
+			);
+		}
+
+		T::DbWeight::get().reads_writes(READ_WEIGHT, WRITE_WEIGHT)
```
apopiak, line 52: You could return the weight of `create_typed_collection`? (plus the version writes)

runtime/testing-basilisk/src/lib.rs
```diff
@@ -919,6 +946,9 @@ construct_runtime!(
 		Marketplace: pallet_marketplace = 109,
 		TransactionPause: pallet_transaction_pause = 110,
 
+		LiquidityMining: pallet_xyk_liquidity_mining = 111,
```
apopiak, line 949: nitpick: won't we have a stable swap liquidity mining? then this should be `XYKLiqduityMining` or similar

runtime/testing-basilisk/src/lib.rs
```diff
@@ -919,6 +946,9 @@ construct_runtime!(
...
+		WarehouseLM: warehouse_liquidity_mining::<Instance1> = 112,
```
apopiak, line 950:
```suggestion
		XYKWarehouseLM: warehouse_liquidity_mining::<XYKLiquidityMiningInstance> = 112,
```
would this be possible? (and would we want to specify the name of the instance on the left?)

runtime/testing-basilisk/src/lib.rs
```diff
@@ -817,6 +817,19 @@ impl pallet_relaychain_info::Config for Runtime {
...
 
+impl pallet_xyk_liquidity_mining::Config for Runtime {
+	type Event = Event;
+	type MultiCurrency = Currencies;
+	type CreateOrigin = EnsureRoot<AccountId>;
+	type PalletId = LMPalletId;
+	type BlockNumberProvider = cumulus_pallet_parachain_system::RelaychainBlockNumberProvider<Runtime>;
+	type NftCollectionId = LiquidityMiningNftCollectionId;
+	type AMM = XYK;
+	type WeightInfo = ();
```
apopiak, line 828: TODO, no?

runtime/common/src/lib.rs
```diff
@@ -276,9 +276,17 @@ parameter_types! {
...
+}
+
+// warehouse pallet liquidity mining
+parameter_types! {
+	pub const WarehouseLMPalletId: PalletId = PalletId(*b"WhouseLm");
+	pub const MinDeposit: Balance = 1;//TODO: tmp value
+	pub const MaxEntriesPerDeposit: u8 = 5; //NOTE: Rebenchmark when this change //TODO: tmp value
+	pub const MaxYieldFarmsPerGlobalFarm: u8 = 5; //TODO: tmp value
+	pub const MinPlannedYieldingPeriods: BlockNumber = 100;  //TODO: tmp value
+	pub const MinTotalFarmRewards: Balance = NATIVE_EXISTENTIAL_DEPOSIT * 1_000; //TODO: tmp value
```
apopiak, line 289: TODO

runtime/testing-basilisk/src/lib.rs
```diff
@@ -858,6 +871,20 @@ impl pallet_multisig::Config for Runtime {
...
+	type PalletId = WarehouseLMPalletId;
+	type MinTotalFarmRewards = MinTotalFarmRewards;
+	type MinPlannedYieldingPeriods = MinPlannedYieldingPeriods;
+	type BlockNumberProvider = cumulus_pallet_parachain_system::RelaychainBlockNumberProvider<Runtime>;
+	type AmmPoolId = AccountId;
+	type MaxFarmEntriesPerDeposit = MaxEntriesPerDeposit;
+	type MaxYieldFarmsPerGlobalFarm = MaxYieldFarmsPerGlobalFarm;
+	type Event = Event;
+}
+
```
apopiak, line 887:
```suggestion
static_assertions::const_assert_ne!(
	<Runtime as warehouse_liquidity_mining::Config<XYKLiquidityMiningInstance>>::MaxFarmEntriesPerDeposit::get(),
	0
);
```
If you like to ensure the config is correct

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,931 @@
...
+		/// Emits `YieldFarmStopped` event when successful.
+		#[pallet::weight(<T as Config>::WeightInfo::stop_yield_farm())]
+		pub fn stop_yield_farm(
+			origin: OriginFor<T>,
+			global_farm_id: GlobalFarmId,
+			asset_pair: AssetPair,
+		) -> DispatchResult {
+			let who = ensure_signed(origin)?;
+
+			let amm_pool_id = T::AMM::get_pair_id(asset_pair);
```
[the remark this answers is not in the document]
martinfridrich: We can't check it here because if AMM is destroyed before farming, the owner must be able to stop and destroy yield farm.

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,931 @@
...
+			let deposit_id = T::LiquidityMiningHandler::deposit_lp_shares(
+				global_farm_id,
+				yield_farm_id,
+				amm_pool_id,
+				shares_amount,
+				Self::get_asset_balance_in_amm_pool,
+			)?;
+
+			Self::lock_lp_tokens(amm_share_token, &who, shares_amount)?;
+			T::NFTHandler::mint_into(&T::NftCollectionId::get(), &deposit_id, &who)?;
```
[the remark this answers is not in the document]
martinfridrich: yes, idea is to query warehouse LM to get deposit details

runtime/testing-basilisk/src/lib.rs
martinfridrich: yes, I was thinking about renaming this but FE is already using in on testnet and my understanding was that we want to deploy it directly to rococo without testnet fist.

runtime/testing-basilisk/src/lib.rs
martinfridrich: This didn't work for me(use instance name instead of `Instance1`) but maybe I did something wrong I'll try it again

pallets/xyk-liquidity-mining/Cargo.toml
```diff
@@ -0,0 +1,65 @@
...
+
+# ORML dependencies
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.28", default-features = false }
+
+# HydraDX dependencies
+warehouse-liquidity-mining = { package="pallet-liquidity-mining", git = "https://github.com/galacticcouncil/warehouse", rev="89bde6d2b8bad2263c554eb3d83b76d877971136", default-features = false }
+hydradx-traits-lm = { package="hydradx-traits", git = "https://github.com/galacticcouncil/warehouse", rev="89bde6d2b8bad2263c554eb3d83b76d877971136", default-features = false }
+
+pallet-nft = { git = "https://github.com/galacticcouncil/warehouse", branch="polkadot-v0.9.28", default-features = false }
+hydradx-traits = { git = "https://github.com/galacticcouncil/warehouse", branch="polkadot-v0.9.28", default-features = false }
```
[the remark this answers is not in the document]
martinfridrich: yes, it's intentional. Other pallets (xyk, nft) are using traits from master and LM is using trait from LM branch. 
In my opinion warehouse LM should be merged first before we updates all the pallets to use the same trait. Otherwise we would have to do all the traits changes in the warehouse LM branch

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,931 @@
...
+//! User is rewarded from the next period after he enters.
+//!
+//! Multiple Incentives
+//!
+//! This feature allow users to redeposit already deposited LP shares to multiple yield farms and
+//! receive incentives from this farms.
+//! Deposit in yield farm is called "farm entry".
+//! Maximal number of redepositing same LP shares is configured by variable: `MaxFarmEntriesPerDeposit`.
+//! Set `MaxFarmEntriesPerDeposit` to `1` to disable multiple incentives scheme. !!!NEVER set this
+//! variable to `0`.
```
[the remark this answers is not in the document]
martinfridrich: yes, this comment should not be here. This value is not configurable in this pallet. I'll add check to warehouse pallet

pallets/xyk-liquidity-mining/src/lib.rs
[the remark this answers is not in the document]
apopiak: Wanna add a comment explaining this?

runtime/testing-basilisk/src/lib.rs
apopiak: ah, this is testing so not as important?

runtime/common/src/lib.rs
martinfridrich: This is for @jak-pan. I don't know what values we want

runtime/testing-basilisk/src/lib.rs
martinfridrich: I would prefer to do `integrity_test` in warehouse

runtime/basilisk/src/lib.rs
```diff
@@ -858,7 +858,7 @@ impl pallet_relaychain_info::Config for Runtime {
 impl pallet_xyk_liquidity_mining::Config for Runtime {
 	type Event = Event;
 	type MultiCurrency = Currencies;
-	type CreateOrigin = EnsureRoot<AccountId>;
+	type CreateOrigin = SuperMajorityTechCommitteeOrRoot; //TODO: check this
```
jak-pan, line 861: I would do just root or full tech comm

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,931 @@
...
+		) -> DispatchResult {
+			T::CreateOrigin::ensure_origin(origin)?;
+
+			let (id, max_reward_per_period) = T::LiquidityMiningHandler::create_global_farm(
+				total_rewards,
+				planned_yielding_periods,
+				blocks_per_period,
+				incentivized_asset,
+				reward_currency,
+				owner.clone(),
```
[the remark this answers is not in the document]
martinfridrich: As we discussed this was resolved by adding comment.

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+	#[cfg_attr(test, derive(PartialEq, Eq))]
+	pub enum Error<T> {
+		/// Nft pallet didn't return an owner.
+		CantFindDepositOwner,
+
+		/// Account balance of amm pool shares is not sufficient.
+		InsufficientAmmSharesBalance,
+
+		/// AMM pool does not exist
+		AmmPoolDoesNotExist,
```
enthusiastmartin, line 164: Since this is for xyk. can all AMM references be named XYK  ?

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+
+		/// NFT representing deposit has been destroyed
+		DepositDestroyed { who: T::AccountId, deposit_id: DepositId },
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Create new liquidity mining program with provided parameters.
+		///
+		/// `owner` account have to have at least `total_rewards` balance. This fund will be
```
enthusiastmartin, line 305:
```suggestion
		/// `owner` account has to have at least `total_rewards` balance. This fund will be
```

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+		/// transferred from `owner` to farm account.
+		///
+		/// The dispatch origin for this call must be `T::CreateOrigin`.
+		/// !!!WARN: `T::CreateOrigin` has power over funds of `owner`'s account and it should be
+		/// configure to trusted origin e.g Sudo or Governance.
```
enthusiastmartin, line 310:
```suggestion
		/// configured to trusted origin e.g Sudo or Governance.
```

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+		/// - `asset_pair`: asset pair identifying yield farm in the global farm.
+		///
+		/// Emits `YieldFarmDestroyed` event when successful.
+		#[pallet::weight(<T as Config>::WeightInfo::destroy_yield_farm())]
+		pub fn destroy_yield_farm(
+			origin: OriginFor<T>,
+			global_farm_id: GlobalFarmId,
+			yield_farm_id: YieldFarmId,
+			asset_pair: AssetPair,
+		) -> DispatchResultWithPostInfo {
```
enthusiastmartin, line 625: Why the return value is different here ? everywhere else is just DispatchResult.

I believe DispatchResult should work ?

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+		/// Emits `SharesRedeposited` event when successful.
+		#[pallet::weight(<T as Config>::WeightInfo::redeposit_lp_shares())]
+		pub fn redeposit_lp_shares(
+			origin: OriginFor<T>,
+			global_farm_id: GlobalFarmId,
+			yield_farm_id: YieldFarmId,
+			asset_pair: AssetPair,
+			deposit_id: DepositId,
+		) -> DispatchResult {
+			let owner = Self::ensure_nft_owner(origin, deposit_id)?;
```
enthusiastmartin, line 719: I would prefer to check signed origin outside the ensure_nft_owner. 

so first check signed origin and pass the account to check the owner instead of origin.

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+				who,
+				amount: shares_amount,
+				lp_token: amm_share_token,
+				deposit_id,
+			});
+
+			Ok(())
+		}
+
+		/// This function create yield farm entry for existing deposit. LP shares are not transferred
```
enthusiastmartin, line 698: to be consistent with other comments - it should be "Create yield ..." without the "This function"

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+				who: owner,
+				claimed,
+				reward_currency,
+				deposit_id,
+			});
+
+			Ok(())
+		}
+
+		/// Withdraw LP shares from liq. mining. with reward claiming if possible.
```
enthusiastmartin, line 784:
```suggestion
		/// Withdraw LP shares from liq. mining with reward claiming if possible.
```

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+		///
+		/// Cases for transfer LP shares and claimed rewards:
```
enthusiastmartin, line 786: this sentence doesn't make much sense to me. 

List of possible cases of transfers of shares and rewards ?

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+			origin: OriginFor<T>,
+			deposit_id: DepositId,
+			yield_farm_id: YieldFarmId,
+			asset_pair: AssetPair,
+		) -> DispatchResult {
+			let owner = Self::ensure_nft_owner(origin, deposit_id)?;
+			let amm_pool_id = Self::ensure_amm(asset_pair)?;
+
+			let global_farm_id = T::LiquidityMiningHandler::get_global_farm_id(deposit_id, yield_farm_id)
+				.ok_or(Error::<T>::DepositDataNotFound)?;
```
enthusiastmartin, line 819: Is this correct error ?

enthusiastmartin (commented):
i havent checked closely tests nor benchmarks to see if they cover the worst case scenarios. If you have any doubts, we can dive in as this requires bit of effort to understand the LM itself.

runtime/testing-basilisk/src/lib.rs
apopiak: It's fine :shrug:

pallets/xyk-liquidity-mining/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+
+		LiquidityMining::<T>::deposit_shares(RawOrigin::Signed(liq_provider.clone()).into(), GLOBAL_FARM_ID, YIELD_FARM_ID, ASSET_PAIR, shares_amount)?;
+
+		set_block_number::<T>(400_000);
+		let liq_provider_bsx_balance = MultiCurrencyOf::<T>::free_balance(BSX, &liq_provider);
+	}: {
+		LiquidityMining::<T>::withdraw_shares(RawOrigin::Signed(liq_provider.clone()).into(), DEPOSIT_ID, YIELD_FARM_ID, ASSET_PAIR)?
+	} verify {
+		assert!(MultiCurrencyOf::<T>::free_balance(BSX, &liq_provider).gt(&liq_provider_bsx_balance));
+	}
```
apopiak, line 419: You probably want to trigger (and then check) the case where it is destroyed?

`withdraw_shares` is quite complex so I'm somewhat skeptical that you've covered the worst case with this short benchmark.

You might even want 2 or 3 for the different scenarios and then compare them in terms of their weight.

pallets/xyk-liquidity-mining/Cargo.toml
```diff
@@ -0,0 +1,64 @@
...
+
+# ORML dependencies
+orml-traits = { git = "https://github.com/open-web3-stack/open-runtime-module-library", branch = "polkadot-v0.9.28", default-features = false }
+
+# HydraDX dependencies
+warehouse-liquidity-mining = { package="pallet-liquidity-mining", git = "https://github.com/galacticcouncil/warehouse", rev="89bde6d2b8bad2263c554eb3d83b76d877971136", default-features = false }
+# This pallet is using 2 different version of triats on purpose because warehouse branch is not merget yet.
+hydradx-traits-lm = { package="hydradx-traits", git = "https://github.com/galacticcouncil/warehouse", rev="89bde6d2b8bad2263c554eb3d83b76d877971136", default-features = false }
+
+hydradx-traits = { git = "https://github.com/galacticcouncil/warehouse", branch="polkadot-v0.9.28", default-features = false }
```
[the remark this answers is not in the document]
martinfridrich: I don't lie it either :) but yes, warehosue LM pallets needs to be merged and other pallets(xyk, nft,...) needs to be updated to use the same trait version.

pallets/xyk-liquidity-mining/src/tests/withdraw_shares.rs
```diff
@@ -0,0 +1,355 @@
...
+
+			assert_noop!(
+				LiquidityMining::withdraw_shares(Origin::signed(CHARLIE), 1, non_known_farm, BSX_KSM_ASSET_PAIR),
+				Error::<Test>::DepositDataNotFound
+			);
+		});
+}
+
+#[test]
+fn deposit_shares_should_fail_when_origin_is_not_signed() {
```
apopiak, line 328:
```suggestion
fn withdraw_shares_should_fail_when_origin_is_not_signed() {
```

pallets/xyk-liquidity-mining/src/lib.rs
```diff
@@ -0,0 +1,927 @@
...
+
+		/// Yield farm can not be found
+		YieldFarmNotFound,
+
+		///Deposit data not found
+		DepositDataNotFound,
+
+		/// Calculated reward to claim is 0.
+		ZeroClaimedRewards,
+	}
```
apopiak, line 180: Missing tests for `CantFindDepositOwner`, `CantGetAmmAssets`, `YieldFarmNotFound`

apopiak (approved):
:+1: for extensive test coverage

2 Points on the benchmarks:
+ I think it would be good to add `verify` clauses for the benchmarks.
+ I suspect the `withdraw_shares` benchmark is not measuring the worst case.

Seems fine to merge once `withdraw_shares` benchmark is checked and/or extended.

pallets/xyk-liquidity-mining/src/lib.rs
martinfridrich: Yes, this error is ok, but I'm open to suggestions for better name. 
Basically `get_global_farm_id()` can return `None` in 2 cases:
1-th - deposit doesn't exists - this case should never happen because NFT must exists at this point (`ensure_nft_owner()`)
2-nd - deposit does exists but there are not data for `yield_farm_id` in the deposit.

pallets/xyk-liquidity-mining/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+
+		initialize_pool::<T>(xyk_caller, ASSET_PAIR.asset_in, ASSET_PAIR.asset_out, 1_000_000 * ONE, Price::from(10))?;
+		let xyk_id = xykpool::Pallet::<T>::pair_account_from_assets(ASSET_PAIR.asset_in, ASSET_PAIR.asset_out);
+		xyk_add_liquidity::<T>(liq_provider.clone(), ASSET_PAIR, 10_000, 1_000_000_000)?;
+
+		lm_create_global_farm::<T>(1_000_000 * ONE, caller.clone(), Perquintill::from_percent(20))?;
+		lm_create_yield_farm::<T>(caller.clone(), GLOBAL_FARM_ID, ASSET_PAIR, FixedU128::one())?;
+
+		lm_deposit_shares::<T>(liq_provider, ASSET_PAIR, 10_000)?;
+		set_block_number::<T>(200_000);
```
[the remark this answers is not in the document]
martinfridrich: yes, this is to change the period and to make sure the farm's RPZ update happens

pallets/xyk-liquidity-mining/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+
+		initialize_pool::<T>(xyk_caller.clone(), ASSET_PAIR.asset_in, ASSET_PAIR.asset_out, 1_000_000 * ONE, Price::from(10))?;
+		let xyk_id = xykpool::Pallet::<T>::pair_account_from_assets(ASSET_PAIR.asset_in, ASSET_PAIR.asset_out);
+		xyk_add_liquidity::<T>(liq_provider.clone(), ASSET_PAIR, 10_000, 1_000_000_000)?;
+
+		lm_create_global_farm::<T>(1_000_000 * ONE, caller.clone(), Perquintill::from_percent(20))?;
+		lm_create_yield_farm::<T>(caller.clone(), GLOBAL_FARM_ID, ASSET_PAIR, FixedU128::one())?;
+
+		lm_deposit_shares::<T>(liq_provider, ASSET_PAIR, 10_000)?;
+		set_block_number::<T>(100_000);
```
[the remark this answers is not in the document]
martinfridrich: yes

pallets/xyk-liquidity-mining/benchmarking/src/lib.rs
```diff
@@ -0,0 +1,461 @@
...
+		//global id: 9, yield id:10
+		lm_create_global_farm::<T>(100_000 * ONE, caller.clone(), Perquintill::from_percent(20))?;
+		lm_create_yield_farm::<T>(caller, 9, ASSET_PAIR, FixedU128::one())?;
+
+		set_block_number::<T>(200_000);
+
+		LiquidityMining::<T>::deposit_shares(RawOrigin::Signed(liq_provider.clone()).into(), GLOBAL_FARM_ID, YIELD_FARM_ID, ASSET_PAIR, shares_amount)?;
+		LiquidityMining::<T>::redeposit_lp_shares(RawOrigin::Signed(liq_provider.clone()).into(), GLOBAL_FARM_ID_2, YIELD_FARM_ID_2, ASSET_PAIR, DEPOSIT_ID)?;
+		LiquidityMining::<T>::redeposit_lp_shares(RawOrigin::Signed(liq_provider.clone()).into(), 5, 6, ASSET_PAIR, DEPOSIT_ID)?;
+		LiquidityMining::<T>::redeposit_lp_shares(RawOrigin::Signed(liq_provider.clone()).into(), 7, 8, ASSET_PAIR, DEPOSIT_ID)?;
```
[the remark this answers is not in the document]
martinfridrich: I'll double check but I think there was a difference when there was only 1 entry in the deposit. I'll add comment.
In the regular deposit it's not necessary because it creates new deposit

pallets/xyk-liquidity-mining/benchmarking/src/lib.rs
martinfridrich: yes, I think I compared case when deposit is destroyed and this one was worse. I'll double check and add comment.

enthusiastmartin (approved):
resolve the TODO and lgtm

lumir-mrkva (approved):
LGTM

lumir-mrkva (approved):
LGTM

PR #528, opened 2022-07-15
feat: enable xyk

enthusiastmartin:
When XYK creates a pool, a pool account is associeted with that pool. 

This account should be excluded from a dusting mechanism - the account cannot be dusted if remaining amounts of assets in pool are below existential deposits. 

This PR introduces a new XYK pallet property which adds pool accounts to the exclusion list. 

Pallet's tests currently dont work with ED and dusting properly - so there was no much sense to add any tests of that.

However an integration test was created to make sure that the integrations with runtime, duster pallet and xyk pallet work properly. 

- updates all warehouse depedencies to latest version 
- enables XYK in runtime ( removes it from filter )


integration-tests/src/xyk.rs
```diff
@@ -0,0 +1,73 @@
...
+	Basilisk::execute_with(|| {
+		//arange
+		assert_ok!(XYK::create_pool(
+			Origin::signed(ALICE.into()),
+			asset_a,
+			asset_b,
+			100 * UNITS,
+			Price::from(2)
+		));
+		assert!(DustRemovalWhitelist::contains(&pair_account));
```
apopiak, line 60:
```suggestion
		assert!(DustRemovalWhitelist::contains(&pair_account(asset_a, asset_b)));
```
nitpick: I would make the pair_account a function

pallets/xyk/src/lib.rs
```diff
@@ -546,6 +552,9 @@ pub mod pallet {
 				<PoolAssets<T>>::remove(&pair_account);
 				<TotalLiquidity<T>>::remove(&pair_account);
 
+				// Ignore the failure, this cant stop liquidity removal
+				let _ = T::NonDustableWhitelistHandler::remove_account(&pair_account);
```
apopiak, line 556: I wonder whether we might want to log this. If this happens something went very wrong.

pallets/duster/src/lib.rs
```diff
@@ -323,3 +323,24 @@ impl<T: Config> Contains<T::AccountId> for DusterWhitelist<T> {
 		AccountBlacklist::<T>::contains_key(t)
 	}
 }
+
+use hydradx_traits::pools::DustRemovalAccountWhitelist;
+
+impl<T: Config> DustRemovalAccountWhitelist<T::AccountId> for Pallet<T> {
+	type Error = DispatchResult;
```
apopiak, line 330: This should really be `type Result`, no?
Or the functions should return `Result<(), Self::Error>`, no?

apopiak (approved):
LGTM

apopiak:
And :+1:, :pray: and thank you for the PR description :smiley:

pallets/duster/src/lib.rs
```diff
@@ -323,3 +323,24 @@ impl<T: Config> Contains<T::AccountId> for DusterWhitelist<T> {
 		AccountBlacklist::<T>::contains_key(t)
 	}
 }
+
+use hydradx_traits::pools::DustRemovalAccountWhitelist;
+
+impl<T: Config> DustRemovalAccountWhitelist<T::AccountId> for Pallet<T> {
```
lumir-mrkva, line 329: it is blacklist or whitelist? and shouldn't this be integrated other way around with handler in duster instead of XYK?

enthusiastmartin: i am aware of the inconsistency. 

it should really be whitelist. orml-tokens uses whitelist, so i would like to be consistent with that .

however, renaming the storage from blacklist to whitelist would require migration - which i dont really want to do right now. 

It can be as part of another task/issue.

enthusiastmartin: and regarding the other suggestion.

i was thinking how to do - and one way i was considering was the other way.

but that's not really duster feature to include "other stuff" and it is reallysomething that XYK want to force its user/consumer to do . 

by doiing this - forcing to configure XYK pallet in a way that pool accounts will be excluded ( or should be excluded ) from dusting , you achieve what XYK pallet needs to have for pool accounts. 

otherwise - it XYK pallet would be used elsewhere - there is no way to make sure what needs to be done with this.

pallets/duster/src/lib.rs
enthusiastmartin: Yes, better.thanks

integration-tests/src/xyk.rs
enthusiastmartin: done

rdovgan (approved):

pallets/xyk/src/lib.rs
enthusiastmartin: added error log in case of error.

lumir-mrkva (approved):

PR #529, opened 2022-07-15
feat: XYK integration tests 

martinfridrich:
Added integration tests to check if pool account is added/removed from duster whitelist

### DIFF
--- a/integration-tests/src/xyk.rs
+++ b/integration-tests/src/xyk.rs
@@ -1,0 +1,73 @@
+#![cfg(test)]
+
+use crate::kusama_test_net::*;
+
+use basilisk_runtime::{DustRemovalWhitelist, Origin, XYK};
+use hydradx_traits::AMM;
+use primitives::{asset::AssetPair, Price};
+use xcm_emulator::TestExt;
+
+use frame_support::{assert_ok, traits::Contains};
+
+#[test]
+fn pair_account_should_be_added_into_whitelist_when_pool_is_created() {
+	TestNet::reset();
+
+	let asset_a = 1;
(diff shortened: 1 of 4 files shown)


integration-tests/src/xyk.rs
```diff
@@ -0,0 +1,73 @@
...
+
+use basilisk_runtime::{DustRemovalWhitelist, Origin, XYK};
+use hydradx_traits::AMM;
+use primitives::{asset::AssetPair, Price};
+use xcm_emulator::TestExt;
+
+use frame_support::{assert_ok, traits::Contains};
+
+#[test]
+fn pair_account_should_be_added_into_whitelist_when_pool_is_created() {
```
dmoka, line 13: great job with the test names, descriptive ones! 👌

dmoka, line 13: great job with the test names, descriptive ones! 👌

runtime/basilisk/src/lib.rs
```diff
@@ -138,7 +138,6 @@ impl Contains<Call> for BaseFilter {
 
 		#[allow(clippy::match_like_matches_macro)]
 		match call {
-			Call::XYK(_) => false,
```
dmoka, line 141: Just for my understanding: This is basically the filter we were talking about, where we can disable features for our runtime? So in this case, we enable XYK as this line is removed...

enthusiastmartin (approved):

runtime/basilisk/src/lib.rs
martinfridrich: yes

PR #544, opened 2022-08-05
feat: use more realistic data in integration tests

dmoka:
## Description

As discussed, done a bit of refactoring in integration tests setup to use more realistic ids and names for the para chains.

## Checklist:


integration-tests/src/kusama_test_net.rs
```diff
@@ -130,7 +133,7 @@ pub fn kusama_ext() -> sp_io::TestExternalities {
 	ext
 }
 
-pub fn hydra_ext() -> sp_io::TestExternalities {
+pub fn karura_ext() -> sp_io::TestExternalities {
 	use basilisk_runtime::{Runtime, System};
```
apopiak, line 137: calling an instantiation of basilisk "karura" seems odd to me :shrug:

dmoka: that's a good point, Alex, I also had doubts about this.

I wonder whether we could use the real karura-runtime as dependency and execute integration tests with it. What do you think, would it be feasible?

apopiak: It's difficult because it would need to depend on the same version of Polkadot and Substrate.
As it would be out of sync with the original anyway we could create a mock version that includes the things we are interested in.

Alternatively I was thinking about writing a multi-process emulator that would allow for decoupling, but that's a longer project.

integration-tests/src/kusama_test_net.rs
```diff
@@ -254,3 +288,318 @@ pub fn expect_basilisk_events(e: Vec<basilisk_runtime::Event>) {
 pub fn vesting_account() -> AccountId {
 	VestingPalletId::get().into_account()
 }
+
+
+//Setting up mock for Karura runtime
+//TODO: once it works properly, extract it to dedicated file/project
```
lumir-mrkva, line 294: mby time to extract it

integration-tests/src/kusama_test_net.rs
lumir-mrkva: yeah, i would call it something different than karura as it is not really karura

lumir-mrkva: but only `OtherParachain` comes to mind 🤦

dmoka: I thought that in the future we are planning to do cross-chain transfers first with Karura, that's why I named it like that. This is also what was recommended by @green-jay.

Of course, happy to rename it to any other one, if we have a better name for it.

lumir-mrkva: pls rename and LGTM, no karura pls

lumir-mrkva (approved):
rename parachain and LGTM

PR #545, opened 2022-08-10
feat: use warehouse currencies

Roznovjak:


lumir-mrkva:
Task linked: [CU-2pjr0p0 Currencies pallet with events](https://app.clickup.com/t/2pjr0p0)

Issue #543, opened 2022-08-05
Bug: Unable to withdraw offer on burned NFT


petersopko:
![image](https://user-images.githubusercontent.com/73316633/183062274-5a708f41-bf17-4dce-abb5-75f7743f7252.png)

![image](https://user-images.githubusercontent.com/73316633/183061996-a9dc3a96-5f93-4121-83c7-0e4ec4f19e9d.png)

### Description
I'm trying to withdraw offer on burned NFT, but there's a `marketplace.ClassOrInstanceUnknown` error
class: 3889603748
instance: 1
https://beta.kodadot.xyz/bsx/gallery/3889603748-1
![image](https://user-images.githubusercontent.com/73316633/183063575-def24b51-d5b4-48fa-8b98-da12e38cc721.png)

### Expected Behavior
I should be able to withdraw this offer.

### Actual Behavior
I'm unable to withdraw offer.

### Steps to Reproduce

1.
2.
3.

### Your Environment

* Toolchain versions: 
* Version used:
* System type: 
* Operating System and version:
* System specs:

Roznovjak:
Thanks for reporting this issue. The issue was fixed in the runtime version 60.

PR #547, opened 2022-08-18
test(marketplace): add missing test

Roznovjak:
Related to #543

### DIFF
--- a/pallets/marketplace/src/tests/withdraw_offer.rs
+++ b/pallets/marketplace/src/tests/withdraw_offer.rs
@@ -136,7 +136,7 @@ fn nft_owner_should_not_have_rights_for_withdrawing_when_nft_is_burned() {
 }
 
 #[test]
-fn withdraw_offer_should_work_when_nft_is_burned() {
+fn nft_offer_maker_should_have_rights_for_withdrawing_when_nft_is_burned() {
 	//Arrange
 	ExtBuilder::default()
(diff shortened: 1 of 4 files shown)


lumir-mrkva:
Task linked: [CU-2qunxcw Marketplace: add missing test](https://app.clickup.com/t/2qunxcw)

pallets/marketplace/src/tests/withdraw_offer.rs
```diff
@@ -134,3 +134,35 @@ fn nft_owner_should_not_have_rights_for_withdrawing_when_nft_is_burned() {
 			);
 		});
 }
+
+#[test]
+fn withdraw_offer_should_work_when_nft_is_burned() {
```
dmoka, line 139: if I am not mistaken, it is the same test as the previous `nft_owner_should_not_have_rights_for_withdrawing_when_nft_is_burned`, and the difference is that here the offer is withdrawn by the offer maker and not by the owner, right?

if so, we could call this test as follows, just to be in line with the previous test name:
`nft_offer_maker_should_have_rights_for_withdrawing_when_nft_is_burned`

dmoka (approved):

pallets/marketplace/src/tests/withdraw_offer.rs
Roznovjak: yes. And in this case there is no error and the offer is actually withdrawn and deposit returned.

dmoka (approved):

PR #549, opened 2022-08-18
test: add integration tests for nft permissions

Roznovjak:


lumir-mrkva:
Task linked: [CU-2kbujbr pallet-nft integration tests for basilisk](https://app.clickup.com/t/2kbujbr)

integration-tests/src/nft_marketplace.rs
```diff
@@ -135,3 +137,105 @@ fn create_class_should_fail_when_relay_chain_location_not_registered() {
...
+fn nft_permissions_should_require_deposit_for_marketplace_class_when_created() {
+	assert!(<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::has_deposit(&ClassType::Marketplace));
+	assert!(!<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::has_deposit(&ClassType::LiquidityMining));
+	assert!(!<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::has_deposit(&ClassType::Redeemable));
+	assert!(!<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::has_deposit(&ClassType::Auction));
+	assert!(!<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::has_deposit(&ClassType::HydraHeads));
+}
+
+#[test]
+fn create_class_should_be_without_deposit_when_created_using_trait() {
```
dmoka, line 196: Maybe we could try to leverage the [test-case](https://crates.io/crates/test-case) crate here, so we could remove some test code duplication.

So the test would look like this as a result:
```
#[test_case(ClassType::Marketplace)]
#[test_case(ClassType::LiquidityMining)]
#[test_case(ClassType::Redeemable)]
#[test_case(ClassType::Auction)]
#[test_case(ClassType::HydraHeads)]
fn create_class_should_be_without_deposit_when_created_using_trait(class_type: ClassType) {
	// Arrange
	TestNet::reset();
	Basilisk::execute_with(|| {
		let initial_balance = Balances::free_balance(&AccountId::from(ALICE));

		// Act
		assert_ok!(
			<NFT as CreateTypedClass<AccountId, ClassId, ClassType>>::create_typed_class(
				AccountId::from(ALICE),
				0,
				class_type
			)
		);

                // Assert
		assert_eq!(initial_balance, Balances::free_balance(&AccountId::from(ALICE)));
	});
}
```

How do you like it?

integration-tests/src/nft_marketplace.rs
```diff
@@ -135,3 +137,105 @@ fn create_class_should_fail_when_relay_chain_location_not_registered() {
...
+	assert!(<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_create(&ClassType::Marketplace));
+	assert!(!<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_create(&ClassType::LiquidityMining));
+	assert!(!<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_create(&ClassType::Redeemable));
+	assert!(!<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_create(&ClassType::Auction));
+	assert!(!<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_create(&ClassType::HydraHeads));
+}
+
+#[test]
+fn nft_permissions_should_allow_marketplace_class_minting() {
+	assert!(<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_mint(&ClassType::Marketplace));
```
dmoka, line 152: Due to the generic parameters, all the function calls in this and the similar tests are a bit noisy.

How about extracting them to separate function like for example for `can_mint`:
```
fn can_mint(class_type: &ClassType) -> bool {
	<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_mint(class_type)
}
```

So at the end you would have something like:
```
#[test]
fn nft_permissions_should_allow_marketplace_class_minting() {
	assert!(can_mint(&ClassType::Marketplace));
	assert!(!can_mint(&ClassType::LiquidityMining));
	assert!(!can_mint(&ClassType::Redeemable));
	assert!(!can_mint(&ClassType::Auction));
	assert!(!can_mint(&ClassType::HydraHeads));
}
```

I believe that it would make the assertions more readable. And also, if something changes in the generic names, etc, then we would only need to change it in only one place.

integration-tests/src/nft_marketplace.rs
Roznovjak: Thanks. I like it. I investigated how to iterate over enum but unfortunately it's not that easy in rust (without modifying the enum and introducing new dependency).

integration-tests/src/nft_marketplace.rs
Roznovjak: You are right. I will update the tests.

Roznovjak: Hmm. I think this change will not be necessary because after applying `test-case` crate as proposed above in your first comment, the issue will be eliminated because there will be just one assert for all enum variants.

dmoka: yes, that's a good point, we could also use the test-case crate to simplify these tests. The only thing is that the tests would be need to split up into two ones, one for negative and one positive case.

Or, we could add a boolean parameter as a test case param, and also introducing two new constants, to make it even more readable:
```
const ALLOW: bool = true;
const NOT_ALLOW: bool = false;

#[test_case(&ClassType::Marketplace, ALLOW)]
#[test_case(&ClassType::LiquidityMining, NOT_ALLOW)]
#[test_case(&ClassType::Redeemable, NOT_ALLOW)]
#[test_case(&ClassType::Auction, NOT_ALLOW)]
#[test_case(&ClassType::HydraHeads, NOT_ALLOW)]
fn nft_permissions_should_allow_marketplace_class_creation(class_type: &ClassType, is_allowed_by_nft_permission: bool) {
	assert_eq!(is_allowed_by_nft_permission, <basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_create(class_type));
}
```

As you wish ;)

integration-tests/src/nft_marketplace.rs
```diff
@@ -135,3 +138,102 @@ fn create_class_should_fail_when_relay_chain_location_not_registered() {
 		);
 	});
 }
+
+#[test_case(ClassType::Marketplace, true ; "marketplace class")]
+#[test_case(ClassType::LiquidityMining, false ; "liquidity mining class")]
+#[test_case(ClassType::Redeemable, false ; "redeemable class")]
+#[test_case(ClassType::Auction, false ; "auction class")]
+#[test_case(ClassType::HydraHeads, false ; "hydra heads class")]
+fn nft_permissions_should_allow_marketplace_class_creation(class_type: ClassType, is_allowed: bool) {
```
dmoka, line 147: Great job with the tests!

Just one thing as stated in one of my other comments, we could make this even more readable by introducing two consts:
```
const ALLOW: bool = true;
const NOT_ALLOW: bool = false;
```

And using these in the `test_case` attributes. By doing this, we could remove the magic booleans. But as you want, I approve the PR anyway. ;)

dmoka (approved):

integration-tests/src/nft_marketplace.rs
```diff
@@ -135,3 +138,102 @@ fn create_class_should_fail_when_relay_chain_location_not_registered() {
...
+		<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::can_create(&class_type)
+	);
+}
+
+#[test_case(ClassType::Marketplace, true ; "marketplace class")]
+#[test_case(ClassType::LiquidityMining, false ; "liquidity mining class")]
+#[test_case(ClassType::Redeemable, false ; "redeemable class")]
+#[test_case(ClassType::Auction, false ; "auction class")]
+#[test_case(ClassType::HydraHeads, false ; "hydra heads class")]
+fn nft_permissions_should_allow_marketplace_class_minting(class_type: ClassType, is_allowed: bool) {
```
dmoka, line 159: The test name does not reveal exactly what we are testing. We are testing more than that.

I am of the opinion that it is one of the exceptional situations where we can easily ditch the `ShouldWhen` template, and just writing:

`test_nft_permission_for_class_minting`

Roznovjak: Ok. I had a hard time trying to come up with a reasonably good test names :)

dmoka: yes, had the same, it is indeed challenging to apply ShouldWhen here. 

This is one of the situations where we should not be dogmatic, and less is more :)

dmoka (approved):

integration-tests/src/nft_marketplace.rs
```diff
@@ -135,3 +137,102 @@ fn create_class_should_fail_when_relay_chain_location_not_registered() {
...
+		<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::has_deposit(&class_type)
+	);
+}
+
+#[test_case(ClassType::Marketplace; "marketplace class")]
+#[test_case(ClassType::LiquidityMining; "liquidity mining class")]
+#[test_case(ClassType::Redeemable; "redeemable class")]
+#[test_case(ClassType::Auction; "auction class")]
+#[test_case(ClassType::HydraHeads; "hydra heads class")]
+fn create_class_should_be_without_deposit_when_created_using_trait(class_type: ClassType) {
```
martinfridrich, line 221: I would like to have tests like this also for the rest of the trait’s functions.

martinfridrich:
This tests are not specific to marketplace, I think they should be in separate NFT integration tests

integration-tests/src/nft_marketplace.rs
Roznovjak: Ok. But in this case we are not talking about integration tests because we are just testing one implementation of the traits and there is no interaction with other pallets involved. And testing the implementation of it implies that it will also work when integrated into other pallets. So I would add tests for this directly to NFT pallet in warehouse repo. @dmoka what do you think?

martinfridrich: In my opinion this should be in the basilisk repo in the nft integration test (not in the marketplace). 
e.g if someone removes classType or changes permissions in the warehouse and updates tests in the warehouse, we may not catch it. It's the same situations as we had with minting without deposits, warehouse repo was updated and merged and we found problem after we tried to use NFT pallet in LM.

In my opinion this is integration test because it test integration of NFT pallet in basilisk chain with basilisk specific requirements.

dmoka: The generic core logic that we are testing here is basically this:

https://github.com/galacticcouncil/warehouse/blob/85257f53b0da5827c103bf4f739474cc1e83d3f7/nft/src/types.rs#L75

But we are actually testing how this pallet_nft is configured/integrated within Basilisk runtime, because we run this:

Because we are calling `<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::...`

... with `ClassType` enum used in the basilisk runtime pallet_nft config implementation

So I think this is fine as an integration test.

dmoka: Just reading @martinfridrich's answer, I agree with that, as it is not only related to marketplace, we could put them into a separate integration test file.

Roznovjak: Hmm. Ok. But I would prefer to have tests next to implementations. Right now we use structs and trait implementations from warehouse crate and test them in Basilisk repo. And we will need to do the same in Hydra repo. In my opinion the best solution would be to not use generic names for structs and traits (e.g. CreateTypedClassWithoutPermissionCheck). Then we don't need to be afraid of some changes in warehouse repo. But I'm totally fine with what you agreed on.

integration-tests/src/nft.rs
```diff
@@ -0,0 +1,251 @@
...
+#[test_case(ClassType::Auction, NOT_ALLOW ; "auction class")]
+#[test_case(ClassType::HydraHeads, NOT_ALLOW ; "hydra heads class")]
+fn test_nft_permission_for_deposit(class_type: ClassType, is_allowed: bool) {
+	assert_eq!(
+		is_allowed,
+		<basilisk_runtime::Runtime as pallet_nft::Config>::Permissions::has_deposit(&class_type)
+	);
+}
+
+// Nonfungibles traits
```
martinfridrich, line 112: hm, now when I'm looking at it you were right. This look more like unit tests, tested code is not using permissions at all. I should look at nft implementation first, sorry.
I think they should be in warehosue as you said, but I'm ok if it's stay here.

martinfridrich (approved):

PR #550, opened 2022-08-23
feat: router execution logic for XYK

dmoka:
Here is the PR for implementing route trades execution logic for XYK

## Checklist:


pallets/xyk/src/lib.rs
```diff
@@ -853,28 +795,13 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
...
-			amount >= T::MinTradingLimit::get(),
-			Error::<T>::InsufficientTradingAmount
-		);
-
-		ensure!(Self::exists(assets), Error::<T>::TokenPoolNotFound);
+		Self::do_validate_buy_common(assets, amount)?;
 
 		let pair_account = Self::get_pair_id(assets);
 
 		let asset_out_reserve = T::Currency::free_balance(assets.asset_out, &pair_account);
```
dmoka, line 802: I wonder how expensive to call `Self::get_pair_id()` and `T::Currency::free_balance()`. Because they are already called in the validate method. If they are expensive to call, I could return them from the `do_validate...()` method.

pallets/xyk/src/lib.rs
```diff
@@ -948,6 +875,192 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
...
+	fn can_create(_asset_a: AssetId, _asset_b: AssetId) -> bool {
+		true
+	}
+}
+
+use hydradx_traits::router::{Executor, ExecutorError, PoolType};
+
+impl<T: Config> Executor<T::AccountId, AssetId, Balance> for Pallet<T> {
+	type Output = Balance;
+	type Error = ();
```
dmoka, line 914: This is still to be done

pallets/xyk/src/lib.rs
```diff
@@ -948,6 +875,220 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
...
+		};
+		Self::do_validate_buy_common(assets, amount_out_with_fee).map_err(|_| ExecutorError::Error(()))?;
+		Self::do_execute_buy(&amm_transfer).map_err(|_| ExecutorError::Error(()))?;
+
+		Ok(())
+	}
+}
+
+impl<T: Config> Pallet<T> {
+	fn do_validate_sell_common(amount: Balance, assets: AssetPair, who: &T::AccountId) -> Result<(), DispatchError> {
```
lumir-mrkva, line 1063: what is the meaning of the `_common` suffix?

pallets/xyk/src/lib.rs
```diff
@@ -948,6 +875,220 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
...
+		pool_type: PoolType<AssetId>,
+		asset_in: AssetId,
+		asset_out: AssetId,
+		amount_in: Self::TradeCalculationResult,
+	) -> Result<Self::TradeCalculationResult, ExecutorError<Self::Error>> {
+		if pool_type != PoolType::XYK {
+			return Err(ExecutorError::NotSupported);
+		}
+
+		let pair = AssetPair { asset_in, asset_out };
```
lumir-mrkva, line 926: the whole trait implementation should be probably in a separate file and use pallet methods instead of pallet using trade execution trait

the relationship should be inverted in my opinion

pallets/xyk/src/lib.rs
```diff
@@ -948,6 +875,220 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, Balance> for Pallet<T> {
...
+			origin: who.clone(),
+			assets,
+			amount: amount_in_without_fee,
+			amount_out: amount_out_without_fee,
+			fee,
+			discount: false,
+			discount_amount: 0,
+		};
+		//TODO: Dani - proper error propagation
+		Self::do_validate_sell_common(amount_in_without_fee, assets, &who).map_err(|_| ExecutorError::Error(()))?;
```
lumir-mrkva, line 1020: just let it use its own error, those errors should be propagated to the executor as well, exectutor shouldn't need any generice 'TradeFailed' error

lumir-mrkva (changes requested):

pallets/xyk/src/lib.rs
dmoka: I just wanted to clearly separate it from the main function `validate_sell`, so it is a common logic used in multiple places. But you are right, having a suffix can cause confusion for the code readers. Probably just `do_validate_sell` alone does the job.

pallets/xyk/src/lib.rs
dmoka: extracting to separate file is clear, good point, just done it!

but I am not sure what you mean by inverting. In the implementation of `Executor`/ (or with new name `TradeExecution`), we already calling the pallet methods, such as `get_pair_id`, `do_validate_sell_common` or `do_execute_sell`.  Or I might miss something...

pallets/xyk/src/lib.rs
dmoka: you are right, let's retain the original error info, there is indeed no need for additional abstraction around them.

On the other hand, I need to keep `ExecutorError` as it is a part of the control flow. 

I wonder what other things I could do for example some random calculation error. Since we want to return/propagate `DispatchError,` (wrapped in ExcecutorError) I would need to use `DispatchError::Other("...")`, so for example:
```
        let amount_in_without_fee = amount_in
            .amount
            .checked_sub(amount_in.fee)
            .ok_or(ExecutorError::Error(DispatchError::Other("Error while calculating sell trade")))?;
````

Is DisptachError::Other a good option here?

Comment (dmoka) [reply pallets/xyk/src/lib.rs]: you are right, let's retain the original error info, there is indeed no need for additional abstraction around them.

On the other hand, I need to keep `ExecutorError` as it is a part of the control flow. 

I wonder what other things I could do for example some random calculation error. Since we want to return/propagate `DispatchError,` (wrapped in ExcecutorError) I would need to use `DispatchError::Other("...")`, so for example:
```
        let amount_in_without_fee = amount_in
            .amount
            .checked_sub(amount_in.fee)
            .ok_or(ExecutorError::Error(DispatchError::Other("Error while calculating sell trade")))?;
````

Is DisptachError::Other a good option here?

File: pallets/xyk/src/lib.rs (hunk 1 above)
Comment (lumir-mrkva) [reply pallets/xyk/src/lib.rs:802]: it is cheap, and this is probably OT for this PR anyway

Comment (jak-pan) [reply pallets/xyk/src/lib.rs:802]: Should be considered as memory read if it's read once b4. I would be interested if the benchmarking correctly assumes this

File: pallets/xyk/src/lib.rs (hunk 4 above)
Comment (lumir-mrkva) [reply pallets/xyk/src/lib.rs]: meanining that common functions should be implemented in XYK without necessity to implement some Executor trait. Executor trait then can be implemented in separate file by just calling whatever it needs to do in XYK

XYK implementation itself shouldn't be contained in executor trait - this was my original issue with @enthusiastmartin implementation that he basically reimplemented XYK for router, you moved XYK impl to router instead

 I would like to keep XYK impl in XYK

Comment (lumir-mrkva) [reply pallets/xyk/src/lib.rs]: for example i would keep `execute_sell` exactly as is and just called it from the `TradeExecution` trait impl

File: pallets/xyk/src/lib.rs (hunk 5 above)
Comment (lumir-mrkva) [reply pallets/xyk/src/lib.rs]: why you need `ExecutorError` for control flow? any error should abort the execution

File: pallets/xyk/src/lib.rs (hunk 4 above)
Comment (lumir-mrkva) [reply pallets/xyk/src/lib.rs]: in a perfect world, the whole xyk/lib would stay the same and route executor imp would be just in another file

File: pallets/xyk/src/lib.rs (hunk 1 above)
Comment (dmoka) [reply pallets/xyk/src/lib.rs:802]: all right, I will leave it like that if it is a memory read, so it should be cheap and fast.

As for benchmarking, I assume that it takes this into account, as the benchmark tests are executed against the runtime with real dependencies. Otherwise, it would be strange, in my opinion.

File: pallets/xyk/src/lib.rs (hunk 4 above)
Comment (dmoka) [reply pallets/xyk/src/lib.rs]: Totally agree with you! We have a session with Martin soon where we will try to simplify and clean the things up.

File: runtime/basilisk/src/benchmarking/route_executor.rs [hunk 6]
```diff
@@ -0,0 +1,175 @@
...
+
+	Ok((main_asset_in, main_asset_out, trades))
+}
+
+runtime_benchmarks! {
+	{ Runtime, pallet_route_executor}
+
+	execute_sell {
+		let n in 1..MAX_NUMBER_OF_TRADES;
+		let assets_and_trades = generate_trades(n).unwrap();
```
enthusiastmartin, line 112: just a nitpick. you can do this

```suggestion
		let (asset_in, asset_out, routes) = generate_trades(n).unwrap();
```

dmoka: this is a good one!

pallets/xyk/src/lib.rs
dmoka: as discussed in the mob-programming session, I leave the error as it is used in the control flow. I mark the comment resolved.

pallets/xyk/src/trade_execution.rs
```diff
@@ -0,0 +1,124 @@
...
+		who: &T::Origin,
+		asset_in: AssetId,
+		asset_out: AssetId,
+		amount_in: Balance,
+	) -> Result<(), ExecutorError<Self::Error>> {
+		if pool_type != PoolType::XYK {
+			return Err(ExecutorError::NotSupported);
+		}
+
+		Self::sell(who.clone(), asset_in, asset_out, amount_in, Balance::zero(), false).map_err(ExecutorError::Error)
```
lumir-mrkva, line 108: origin shouldn't be cloned here, but outside of the function

lumir-mrkva (approved):

Issue #553, opened 2022-08-31
Bug: Cannot add liquidity to XYK if limit is set higher than balance


jak-pan:
https://github.com/galacticcouncil/Basilisk-node/blob/7a1bddd4670c7dcc4b14489b8b5799685f91246d/pallets/xyk/src/lib.rs#L410-L413

This line specifically forbids you to add to the pool even though the tx would have passed if you set high max limit which should not be the case. I would personally remove this condition.

Roznovjak:
> https://github.com/galacticcouncil/Basilisk-node/blob/7a1bddd4670c7dcc4b14489b8b5799685f91246d/pallets/xyk/src/lib.rs#L410-L413
> 
> This line specifically forbids you to add to the pool even though the tx would have passed if you set high max limit which should not be the case. I would personally remove this condition.

What about replacing it with `ensure!(T::Currency::free_balance(asset_b, &who) >= amount_b, ...` where `amount_b` is the actual amount that will be deducted from the account? To keep the errors consistent...

enthusiastmartin:
> https://github.com/galacticcouncil/Basilisk-node/blob/7a1bddd4670c7dcc4b14489b8b5799685f91246d/pallets/xyk/src/lib.rs#L410-L413
> 
> This line specifically forbids you to add to the pool even though the tx would have passed if you set high max limit which should not be the case. I would personally remove this condition.

yes.agreed. 

this check can be removed.  

richard's suggestions can be added instead.

PR #555, opened 2022-09-02
fix(xyk): add liquidity checks calculated amount instead of limit

Roznovjak:
Fixes: #553


lumir-mrkva:
Task linked: [CU-2rx7dtm XYK - test correct asset_b amount in add_liquidity](https://app.clickup.com/t/2rx7dtm)

pallets/xyk/src/tests/liquidity.rs
```diff
@@ -185,6 +185,34 @@ fn add_liquidity_as_another_user_should_work() {
 	});
 }
 
+#[test]
+fn add_liquidity_should_work_when_limit_exceeds_account_balance() {
```
enthusiastmartin, line 189: this does not sound right.  if a limit is exceeded - stuff should not work. 

perhaps just better wording here. 

i would like to see a test where it fails.

Roznovjak: Renamed.
The test where it fails is [here](https://github.com/galacticcouncil/Basilisk-node/blob/7a1bddd4670c7dcc4b14489b8b5799685f91246d/pallets/xyk/src/tests/liquidity.rs#L395)

pallets/xyk/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = 'pallet-xyk'
-version = '5.0.1'
+version = '5.0.2'
```
lumir-mrkva, line 3: 5.1.0

lumir-mrkva (changes requested):
please bump runtime version

enthusiastmartin (approved):

jak-pan (approved):

PR #559, opened 2022-09-09
feat: royalty type

### LINKED ISSUE #556: Change royalty bond to be more granular
Currently we only allow 1% increments in royalty settings I would expect it to be at least 0.1% increments if not 0.01% (this would require u8 -> u16 type change)

https://github.com/galacticcouncil/Basilisk-node/blob/7a1bddd4670c7dcc4b14489b8b5799685f91246d/pallets/marketplace/src/lib.rs#L301

Roznovjak:
Fixes #556

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5932,6 +5932,7 @@ dependencies = [
  "primitives",
  "scale-info",
  "serde",
+ "sp-arithmetic",
  "sp-core",
  "sp-io",
  "sp-runtime",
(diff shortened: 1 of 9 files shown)


jak-pan (approved):

jak-pan:
LGTM -> need to check with Koda

vikiival:
Jup easy fix for us, we just need a release to handle it

PR #560, opened 2022-09-19
fix!: remove pallet deps from primitives crate

apopiak:
primitives crates should avoid including complex dependencies where possible


dmoka (approved):
LGTM.

I wonder why they were added in the first place, as they were not used at all. Might be some leftovers...

enthusiastmartin (approved):

PR #563, opened 2022-09-23
fix(lbp): validate calculated asset_in amount

Roznovjak:


lumir-mrkva (approved):

pallets/lbp/src/tests.rs
```diff
@@ -2372,6 +2372,59 @@ fn buy_should_work() {
...
+				amount: 5_560_301,
+				buy_price: 10_000_000,
+				fee_asset: KUSD,
+				fee_amount: 20_000,
+			}
+			.into(),
+		]);
+	});
+}
+
```
dmoka, line 2427: I think we should also add negative test cases triggering the `InsufficientAssetBalance` error, both when we buy accumulated and distributed assets.

dmoka (approved):
Apart from some missing test cases, it looks fine!

If you can add those, that would be great, then we can merge it back.

pallets/lbp/src/lib.rs
```diff
@@ -1157,6 +1153,11 @@ impl<T: Config> AMM<T::AccountId, AssetId, AssetPair, BalanceOf<T>> for Pallet<T
 				Error::<T>::MaxInRatioExceeded
 			);
 
+			ensure!(
```
dmoka, line 1156: I think we should also add negative test cases triggering the InsufficientAssetBalance error, both when we buy accumulated and distributed assets.

dmoka (approved):

pallets/lbp/src/tests.rs
Roznovjak: I have just added some two missing cases. Please check the latest commit.

PR #565, opened 2022-10-03
feat: implement router for LBP

dmoka:
## Description

Added implementation and integration tests for router in the context of LBP

## Checklist:


pallets/lbp/src/trade_execution.rs
```diff
@@ -0,0 +1,150 @@
...
+			let calculated_in = hydra_dx_math::lbp::calculate_in_given_out(
+				asset_in_reserve,
+				asset_out_reserve,
+				weight_in,
+				weight_out,
+				amount_out_plus_fee,
+			)
+			.map_err(|_| ExecutorError::Error(Error::<T>::Overflow.into()))?;
+
+			Ok(calculated_in) //TODO: Double check with someone if this is correct
```
dmoka, line 104: I am not sure about this. Although it seems it works with this according to the integration tests, I expected to remove the fee from this amount, like in `calculate_sell` line 46, as the pool is responsible for sending the fee to the fee-collector

integration-tests/src/router.rs
```diff
@@ -1,623 +1,1325 @@
...
+					BSX,
+					AUSD,
+					amount_to_sell * UNITS,
+					limit,
+					trades
+				),
+				pallet_xyk::Error::<basilisk_runtime::Runtime>::TokenPoolNotFound
+			);
+
+			assert_trader_bsx_balance(BOB_INITIAL_BSX_BALANCE);
```
Roznovjak, line 303: It's not necessary to test anything after `assert_noop`.

dmoka: I am doing these checks in the tests because in a route we can do multiple trades, and for each trade there is a currency transfer happening. 

If one of them fails I want to be sure that everything is rolled back.

I could introduce some helper methods to make these assertion intents more explicit, for exa,ple: `asset_that_trader_bas_balance_has_not_changed`

integration-tests/Cargo.toml
```diff
@@ -107,9 +107,11 @@ polkadot-runtime-parachains = { git = "https://github.com/paritytech/polkadot",
 basilisk-runtime = { path = "../runtime/basilisk", default-features = false}
 kusama-runtime = { git = "https://github.com/paritytech/polkadot", branch = "release-v0.9.17" }
 
+cumulus-test-relay-sproof-builder = { git = "https://github.com/paritytech/cumulus", branch = "polkadot-v0.9.26" }
```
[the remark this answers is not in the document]
dmoka: good catch!

pallets/lbp/src/trade_execution.rs
lumir-mrkva: checks out, looks the same as lbp impl, pls remove TODO

lumir-mrkva (approved):
i would prefer instead of rewriting existing tests to include LBP write new ones specifically for it

Issue #566, opened 2022-10-05
change: xyk - check max out/in ratio in sell/buy


enthusiastmartin:
In XYK, currently only the specified amount is checked against max allowed in/out reserve ratio.

eg. 
in sell - only amount_in is checked to be <= reserve / MaxInRatio
in buy - only amount_out is checked to be <= reserve/ MaxOutRatio.

It might be worth to check the amount out in sell, and amount in in buy as well. Similar to what we do in LBP or Omnipool. 

@jak-pan wdyt ?

jak-pan:
We can, but this is not as critical because there is no reason to expect math errors like in omnipool. So it's not critical.

PR #570, opened 2022-10-25
feat(xyk): test max in/out ratios

Roznovjak:
Fixes: #566


Roznovjak:
I had to change the value of `MaxOutRatio` in one test because I believe it's not possible to reach the limit and test it if the ratios are set to `1/3`.

lumir-mrkva (approved):

lumir-mrkva (approved):

PR #571, opened 2022-10-31
chore: upgrade to polkadot-v0.9.29

enthusiastmartin:
Changes:
 - Weight type changed from u64 to a struct. 
 - math lib contains changes in lbp math (rounding fixed by colin) and improved pow accuracy
 - removed Exchange pallet
 - removed LBP and XYK rpcs
 - Weights updated


runtime/basilisk/src/lib.rs
```diff
@@ -1045,7 +1032,7 @@ pub type Executive = frame_executive::Executive<
 	Block,
 	frame_system::ChainContext<Runtime>,
 	Runtime,
-	AllPalletsReversedWithSystemFirst,
+	AllPalletsWithSystem,
```
lumir-mrkva, line 1035: why did this change and are youaware of the impact of this?

enthusiastmartin: > @enthusiastmartin why did this change? no change in `construct_runtime!` is required?

AllPalletsReversedWithSystemFirst is deprecated in favor of this.

lumir-mrkva: i am aware of it, but it also changes orders in which are pallets loaded

lumir-mrkva (approved):

PR #579, opened 2022-11-07
chore: remove old migrations

martinfridrich:
All old migrations was removed.


runtime/basilisk/src/migrations.rs
```diff
@@ -54,10 +46,6 @@ impl OnRuntimeUpgrade for OnRuntimeUpgradeMigration {
 		pallet_nft::migration::v1::post_migrate::<Runtime>();
```
lumir-mrkva, line 46: @martinfridrich can we remove this as well?

lumir-mrkva (approved):

PR #583, opened 2022-11-11
fix: base_path uses tmp dir

Roznovjak:
Fix: `base_path()` uses temp dir when `-tmp` param is used

### DIFF
--- a/node/src/command.rs
+++ b/node/src/command.rs
@@ -443,9 +443,10 @@ impl CliConfiguration<Self> for RelayChainCli {
 	}
 
 	fn base_path(&self) -> Result<Option<BasePath>> {
-		self.shared_params()
-			.base_path()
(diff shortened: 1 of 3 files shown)


node/src/command.rs
```diff
@@ -443,9 +443,10 @@ impl CliConfiguration<Self> for RelayChainCli {
...
 
 	fn base_path(&self) -> Result<Option<BasePath>> {
-		self.shared_params()
-			.base_path()
-			.or_else(|_| Ok(self.base_path.clone().map(Into::into)))
+		Ok(self
+			.shared_params()
+			.base_path()?
+			.or_else(|| self.base_path.clone().map(Into::into)))
 	}
```
enthusiastmartin, line 450: why i did not do that like that in first place? hm, my bad.

enthusiastmartin (approved):

PR #588, opened 2022-12-07
fix: csp docs

jak-pan:


Roznovjak:
> 

No, it's not possible.

PR #591, opened 2023-01-09
chore: use latest warehouse

martinfridrich:
Update to use latest warehouse.
`asset-registry` is kept back intentionally.

### DIFF
--- a/pallets/xyk-liquidity-mining/src/tests/withdraw_shares.rs
+++ b/pallets/xyk-liquidity-mining/src/tests/withdraw_shares.rs
@@ -387,3 +387,88 @@ fn withdraw_shares_should_fail_when_nft_owner_is_not_found() {
 			);
 		});
 }
+
+#[test]
+fn withdraw_shares_should_when_yield_farm_is_not_claimable() {
+	ExtBuilder::default()
+		.with_endowed_accounts(vec![
+			(BOB, BSX, 1_000_000 * ONE),
+			(CHARLIE, BSX_KSM_SHARE_ID, 200 * ONE),
+		])
+		.with_amm_pool(BSX_KSM_AMM, BSX_KSM_SHARE_ID, BSX_KSM_ASSET_PAIR)
+		.with_global_farm(
+			500_000 * ONE,
+			20_000,
+			10,
+			BSX,
+			BSX,
+			ALICE,
+			Perquintill::from_percent(1),
+			ONE,
+			One::one(),
+		)
+		.with_yield_farm(ALICE, 1, One::one(), None, BSX_KSM_ASSET_PAIR)
+		.with_deposit(CHARLIE, 1, 2, BSX_KSM_ASSET_PAIR, 100 * ONE)
+		.build()
+		.execute_with(|| {
+			//Arrange
+			set_block_number(1_000);
+
+			let charlie_lp_token_balance = Tokens::free_balance(BSX_KSM_SHARE_ID, &CHARLIE);
+			pretty_assertions::assert_eq!(
+				Tokens::free_balance(BSX_KSM_SHARE_ID, &LiquidityMining::account_id()),
+				100 * ONE
+			);
+
+			assert_ok!(LiquidityMining::stop_yield_farm(
+				Origin::signed(ALICE),
+				1,
+				BSX_KSM_ASSET_PAIR
+			));
+
+			//Act
+			assert_ok!(LiquidityMining::withdraw_shares(
+				Origin::signed(CHARLIE),
+				1,
+				2,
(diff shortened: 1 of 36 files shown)


mrq1911:
> My clippy fails with this error:
> 
> ```
> error: the `Err`-variant returned from this function is very large
>   --> runtime/basilisk/src/benchmarking/route_executor.rs:39:63
>    |
> 39 | pub fn register_asset_with_name(name_as_bye_string: &[u8]) -> Result<AssetId, BenchmarkError> {
>    |                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the `Err`-variant is at least 128 bytes
>    |
>    = help: try reducing the size of `frame_benchmarking::BenchmarkError`, for example by boxing large elements or replacing it with `Box<frame_benchmarking::BenchmarkError>`
>    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
>    = note: `-D clippy::result-large-err` implied by `-D warnings`
> ```
> 
> I needed to add the following line to `runtime/basilisk/src/benchmarking/route_executor.rs` to fix it
> 
> ```
> #![allow(clippy::result_large_err)]
> ```

fix pls

pallets/xyk-liquidity-mining/src/tests/withdraw_shares.rs
```diff
@@ -387,3 +387,88 @@ fn withdraw_shares_should_fail_when_nft_owner_is_not_found() {
 			);
 		});
 }
+
+#[test]
+fn withdraw_shares_should_when_yield_farm_is_not_claimable() {
```
Roznovjak, line 392: Please update the name of the test. The expected behavior is missing.

runtime/common/src/lib.rs
```diff
@@ -304,3 +304,12 @@ parameter_types! {
 	pub DepositFactor: Balance = deposit(0, 32);
 	pub const MaxSignatories: u16 = 100;
 }
+
+//pallet uniques
+parameter_types! {
+	pub const CollectionDeposit: Balance = 0;
+	pub const ItemDeposit: Balance = 0;
+	pub const KeyLimit: u32 = 256;	// Max 256 bytes per key
+	pub const ValueLimit: u32 = 1024;	// Max 1024 bytes per value
```
Roznovjak, line 313: We still have these four values defined in the normal and testing runtime. Please remove them from there.

runtime/common/src/lib.rs
```diff
@@ -304,3 +304,12 @@ parameter_types! {
...
+	pub const UniquesStringLimit: u32 = 72;
```
Roznovjak, line 314: We can use `pub const UniquesStringLimit: u32 = primitives::UniquesStringLimit;` and remove `primitives::` prefix from the runtimes and use just `UniquesStringLimit` (instead of `primitives::UniquesStringLimit`)

runtime/basilisk/src/benchmarking/route_executor.rs
```diff
@@ -37,6 +37,7 @@ pub const UNITS: Balance = 100_000_000_000;
 const MAX_NUMBER_OF_TRADES: u32 = 5;
 
 pub fn register_asset_with_name(name_as_bye_string: &[u8]) -> Result<AssetId, BenchmarkError> {
+	#![allow(clippy::result_large_err)]
```
Roznovjak, line 40: One small thing. I got the error two times. From `register_asset_with_name` and also from `generate_trades`. But this line catches both of them because the attribute is used with `!` after `#`. Then it is applied to a crate. Can we place it at the beginning of the file? Thanks. Some [info](https://stackoverflow.com/questions/25877285/how-to-disable-unused-code-warnings-in-rust)

runtime/common/src/lib.rs
```diff
@@ -304,12 +304,3 @@ parameter_types! {
...
 	pub const MaxSignatories: u16 = 100;
 }
-
-//pallet uniques
-parameter_types! {
-	pub const CollectionDeposit: Balance = 0;
-	pub const ItemDeposit: Balance = 0;
-	pub const KeyLimit: u32 = 256;	// Max 256 bytes per key
-	pub const ValueLimit: u32 = 1024;	// Max 1024 bytes per value
-	pub const UniquesStringLimit: u32 = 72;
```
Roznovjak: I think it's better to leave it here and remove it from runtimes to have it defined just once.

Roznovjak:
Probably not related to this PR, but do we need `BlockNumberProvider` in the `xyk-liquidity-mining::Config`?

Roznovjak (approved):

PR #592, opened 2023-01-15
feat: English auctions

vgantchev:
This PR adds NFT auctions with the first supported type: English auctions.

The PR is a continuation of the ongoing work on Auctions as witnessed by https://github.com/galacticcouncil/Basilisk-node/pull/168


vgantchev:
FYI @kuchta @vikiival @yangwao

This feat will become available in `testing-basilisk` runtime once merged

pallets/auctions/Cargo.toml
```diff
@@ -0,0 +1,53 @@
+[package]
+name = "pallet-auctions"
+version = "1.0.0"
+authors = ["GalacticCoucil"]
+description = "NFT Auctions"
+edition = "2018"
```
green-jay, line 6:
```suggestion
edition = "2021"
```

pallets/auctions/src/lib.rs
```diff
@@ -0,0 +1,807 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		///
+		/// Creates a new auction for a given Auction type
+		///
+		/// - calls the create() implementation on the given Auction type
+		///
+		/// Parameters:
+		/// - `auction` - Auction data (common and specific for the particular auction type)
```
green-jay, line 336: not very helpful, all params should be described

runtime/src/weights/auctions.rs
```diff
@@ -0,0 +1,83 @@
...
+};
+use sp_std::marker::PhantomData;
+
+use pallet_auctions::weights::WeightInfo;
+
+pub struct BasiliskWeight<T>(PhantomData<T>);
+
+impl<T: frame_system::Config> WeightInfo for BasiliskWeight<T> {
+	fn create_auction() -> Weight {
+		(32_486_000 as Weight)
```
green-jay, line 54: which weights are correct? should be unified with the other ones

runtime/basilisk/src/lib.rs
```diff
@@ -837,6 +838,29 @@ impl pallet_uniques::Config for Runtime {
 	type Helper = ();
 }
 
+parameter_types! {
+	pub const AuctionsStringLimit: u32 = 128; // limit length of auction name
```
green-jay, line 842: this is perhaps too much, we could unify with nft metadata limit (was 72 iirc)

pallets/auctions/src/mock.rs
```diff
@@ -0,0 +1,252 @@
...
+	}
+);
+
+/// Balance of an account.
+pub type Balance = u128;
+
+pub const ALICE: AccountId = AccountId::new([1u8; 32]);
+pub const BOB: AccountId = AccountId::new([2u8; 32]);
+pub const CHARLIE: AccountId = AccountId::new([3u8; 32]);
+pub const BSX: Balance = 100_000_000_000;
```
green-jay, line 67:
```suggestion
pub const BSX: Balance = 1_000_000_000_000;
```

pallets/auctions/src/weights.rs
```diff
@@ -0,0 +1,111 @@
+// This file is part of Basilisk-node.
```
green-jay, line 1: this file seems obsolete

pallets/auctions/src/types/english.rs
```diff
@@ -0,0 +1,169 @@
...
+use super::*;
+
+///
+/// Implementation of EnglishAuction
+///
+impl<T: Config> NftAuction<T::AccountId, T::AuctionId, BalanceOf<T>, Auction<T>, Bid<T>> for EnglishAuction<T> {
+	///
+	/// Creates an English Auction
+	///
+	#[require_transactional]
```
green-jay, line 27: aren't all fns transactional by default?

green-jay (changes requested):
direct dependency on pallet-nft should be also deprecated in favour of using nft traits, similar to LM or Omnipool

pallets/auctions/src/types/english.rs
vgantchev: @apopiak can you advise on this? I added the require_transactional upon your request in the original PR.

apopiak: extinsics are transactional by default, yes, but I think it makes sense to still tag functions that require this as the function could be called from an extrinsic that opts out of transactionality or could be called from `on_initialize` which is not transactional by default IIRC.

green-jay: makes sense, esp. with helper fns

Issue #595, opened 2023-01-23
update to latest registry + integration tests


enthusiastmartin:
integration test for XYK pools is required to test that sequential ids are still work correctly with new registry where it is offset-ed by some range.

PR #597, opened 2023-01-31
feat: use latest asset registry

Roznovjak:
Fixes: #595

### DIFF
--- a/integration-tests/src/kusama_test_net.rs
+++ b/integration-tests/src/kusama_test_net.rs
@@ -191,6 +191,12 @@ pub fn basilisk_ext() -> sp_io::TestExternalities {
 			(b"MOVR".to_vec(), 1_000u128),
 			(b"NEW_BOOTSRAPPED_TOKEN".to_vec(), 1_000u128),
 		],
+		asset_ids: vec![
+			(b"KSM".to_vec(), 1_000_000u128, 1),
+			(b"aUSD".to_vec(), 1_000u128, 2),
+			(b"MOVR".to_vec(), 1_000u128, 3),
+			(b"NEW_BOOTSRAPPED_TOKEN".to_vec(), 1_000u128, 4),
+		],
 		native_asset_name: b"BSX".to_vec(),
 		native_existential_deposit: existential_deposit,
 	}
@@ -269,6 +275,7 @@ pub fn other_parachain_ext() -> sp_io::TestExternalities {
 
 	pallet_asset_registry::GenesisConfig::<ParachainRuntime> {
 		asset_names: vec![(b"AUSD".to_vec(), 1_000_000u128)],
+		asset_ids: vec![(b"AUSD".to_vec(), 1_000_000u128, 1)],
 		native_asset_name: b"BSX".to_vec(),
 		native_existential_deposit: existential_deposit,
 	}
--- a/integration-tests/src/kusama_test_net.rs
+++ b/integration-tests/src/kusama_test_net.rs
@@ -185,17 +185,11 @@ pub fn basilisk_ext() -> sp_io::TestExternalities {
 	.unwrap();
 
 	pallet_asset_registry::GenesisConfig::<Runtime> {
-		asset_names: vec![
-			(b"KSM".to_vec(), 1_000_000u128),
-			(b"aUSD".to_vec(), 1_000u128),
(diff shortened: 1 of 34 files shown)


enthusiastmartin:
can you somehow test scenario where xyk pool is created prior to the feature which offsets the ids ? 

and that xyk still works correctly ? i don't think xyk constructs the share asset id - only in create pool. but it would be worth to try it.

Roznovjak:
> can you somehow test scenario where xyk pool is created prior to the feature which offsets the ids ?
> 
> and that xyk still works correctly ? i don't think xyk constructs the share asset id - only in create pool. but it would be worth to try it.

I think we can simulate this by creating a pool with given ID which is within the reserved space, and after that create a new pool without specifying ID. What do you think?

Roznovjak:
> > can you somehow test scenario where xyk pool is created prior to the feature which offsets the ids ?
> > and that xyk still works correctly ? i don't think xyk constructs the share asset id - only in create pool. but it would be worth to try it.
> 
> I think we can simulate this by creating a pool with given ID which is within the reserved space, and after that create a new pool without specifying ID. What do you think?

@enthusiastmartin what do you think about it?

enthusiastmartin:
> 

i am sure how you'd create a pool with given id. 

but you could try to create a pool when the offset is set to 0 first, and then change it to 1_000_000 and continue.

Roznovjak:
> > 
> 
> i am sure how you'd create a pool with given id.
> 
> but you could try to create a pool when the offset is set to 0 first, and then change it to 1_000_000 and continue.

I don't think it's possible to do it in the integration tests because we don't use a mock runtime, so we can't use the builder pattern.

integration-tests/src/kusama_test_net.rs
```diff
@@ -191,6 +191,12 @@ pub fn basilisk_ext() -> sp_io::TestExternalities {
 			(b"MOVR".to_vec(), 1_000u128),
 			(b"NEW_BOOTSRAPPED_TOKEN".to_vec(), 1_000u128),
 		],
+		asset_ids: vec![
+			(b"KSM".to_vec(), 1_000_000u128, 1),
+			(b"aUSD".to_vec(), 1_000u128, 2),
+			(b"MOVR".to_vec(), 1_000u128, 3),
+			(b"NEW_BOOTSRAPPED_TOKEN".to_vec(), 1_000u128, 4),
+		],
```
apopiak, line 199: Same comment as in another PR for hydradx from Martin: Tokens should be in `asset_names` OR `asset_ids`, otherwise it will be registered twice. Sorry for making that non-obvious in my changes to the pallet.

enthusiastmartin:
@Roznovjak it looks ok to me. however i would like to see a test like mentioned above. 

where there is a pool with share token not offseted ( like created before this upgrade)

i believe you can do that in the actual xyk tests ( not integration tests) since it uses pallet-asset-registry in the tests:
1. create a pool
2. manually change the id of share asset id in the storage to simulate such scenario ( in the registry as well as in the xyk storage )
3. and make sure it all works with that share asset id.

integration-tests/src/kusama_test_net.rs
Roznovjak: Ok. I'm going to update the pallet so we can include that change in this PR.

pallets/xyk-liquidity-mining/benchmarking/src/mock.rs
```diff
@@ -344,7 +346,7 @@ impl ExtBuilder {
 		.unwrap();
 
 		pallet_asset_registry::GenesisConfig::<Test> {
-			asset_names: vec![(b"KSM".to_vec(), 1_000), (b"DOT".to_vec(), 1_000)],
+			registered_assets: vec![(b"KSM".to_vec(), 1_000, Some(1)), (b"DOT".to_vec(), 1_000, Some(2))],
```
apopiak, line 349:
```suggestion
			registered_assets: vec![(b"KSM".to_vec(), 1_000, Some(KSM)), (b"DOT".to_vec(), 1_000, Some(2))],
```
probably introduce `pub const DOT: AssetId = 2;` and use that for the initialization here, too?

apopiak (approved):
LGTM, thanks for doing the update :pray:

PR #604, opened 2023-03-01
feat: reduce enactment period

green-jay:
+ separate vote locking from enactment period
+ shorten enactment period from 6 days to 12 hours


runtime/common/src/lib.rs
```diff
@@ -187,7 +187,9 @@ parameter_types! {
 	pub const VotingPeriod: BlockNumber = 3 * DAYS;
 	pub const FastTrackVotingPeriod: BlockNumber = 3 * HOURS;
 	pub const MinimumDeposit: Balance = 1000 * DOLLARS;
-	pub const EnactmentPeriod: BlockNumber = 6 * DAYS;
+	pub const EnactmentPeriod: BlockNumber = 12 * HOURS;
+	// Make sure VoteLockingPeriod > EnactmentPeriod
+	pub const VoteLockingPeriod: BlockNumber = 3 * DAYS;
```
jak-pan, line 192: shouldn't this be 6?

PR #608, opened 2023-04-03
feat: Introduce Tinkernet multisig XCM configs

arrudagates:
This PR adds the XCM configs necessary to allow Tinkernet multisigs to transact on Basilisk, the following are the added configs:

### In Barrier
```rust
invarch_xcm_builder::AllowPaidTinkernetMultisig,
```
This barrier first checks if the XCM origin is a Tinkernet multisig matching the following MultiLocation pattern
```rust
MultiLocation {
    parents: _,
    interior: Junctions::X3(
        Junction::Parachain(2125),
        Junction::PalletInstance(71),
        Junction::GeneralIndex(_)
    )
}
```
And then it checks if the instructions in the message contain a DescendOrigin followed by WithdrawAsset and BuyExecution to ensure the multisig is paying the proper fees for execution in Basilisk.

### In LocationToAccountId
```rust
invarch_xcm_builder::TinkernetMultisigAsAccountId<AccountId>,
```
This MultiLocation converter derives an AccountId for locations matching the one referred above in the barrier, it does so by using the same exact derivation function used in Tinkernet, this is important to make sure the multisigs maintain the same address across all chains, providing the best UX possible.
Because this derivation happens in Basilisk, it means the whole process is trustless and removes any possibility of account impersonation!

### In XcmOriginToCallOrigin
```rust
invarch_xcm_builder::DeriveOriginFromTinkernetMultisig<Origin>,
```
Same as explained above, except here we need to derive the AccountId and wrap it with a RawOrigin::Signed origin so this account can transact in Basilisk.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1205,7 +1205,7 @@ dependencies = [
 
 [[package]]
 name = "common-runtime"
-version = "2.3.7"
+version = "2.3.8"
 dependencies = [
  "cumulus-pallet-xcmp-queue",
  "frame-support",
@@ -3428,8 +3428,8 @@ dependencies = [
 
 [[package]]
 name = "hydra-dx-math"
-version = "5.1.4"
-source = "git+https://github.com/galacticcouncil/HydraDX-math?rev=a0cd48c26dca53e7e351434276c572d70ffd0b72#a0cd48c26dca53e7e351434276c572d70ffd0b72"
+version = "7.0.0"
+source = "git+https://github.com/galacticcouncil/HydraDX-math?rev=84632c3c7b71e16d53240e4bdc2adc693c941916#84632c3c7b71e16d53240e4bdc2adc693c941916"
 dependencies = [
  "fixed",
  "num-traits",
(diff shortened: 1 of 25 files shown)


vgantchev:
@arrudagates can you pls bump versions at these places:

- [ ] https://github.com/arrudagates/Basilisk-node/blob/master/runtime/basilisk/Cargo.toml#L3
- [ ] https://github.com/arrudagates/Basilisk-node/blob/master/runtime/basilisk/src/lib.rs#L114
- [ ] https://github.com/arrudagates/Basilisk-node/blob/master/runtime/testing-basilisk/Cargo.toml#L3
- [ ] https://github.com/arrudagates/Basilisk-node/blob/master/runtime/testing-basilisk/src/lib.rs#L132

vgantchev:
can we expect that the multisig dependencies will be supported in the future and upgraded to latest polkadot versions which will allow us to upgrade them simultaneously?

arrudagates:
> @arrudagates can you pls bump versions at these places:

Done!

mrq1911:
> Is there something wrong with GitHub actions? I see it gives out a "No space left" error.
> 
> https://github.com/galacticcouncil/Basilisk-node/actions/runs/4674590667/jobs/8311817703#step:9:869

looks like our CI script cannot run in the github infra (which is used for the PR from fork)

jak-pan:
> > can we expect that the multisig dependencies will be supported in the future and upgraded to latest polkadot versions which will allow us to upgrade them simultaneously?
> 
> Yes, we will implement automatic updates creating branches for every new Polkadot release, similarly to how Parity handles automatic updates of the substrate runtime templates.
> 
> In the longer term I do believe Parity will refactor the Substrate versioning workflow allowing for all the crates to be published to crates.io with semver, which would then mean once the teams migrate from git versions to semver versions there won't be dependency conflicts anymore and you'll be able to simply add our dependency with version `*` and cargo will automagically settle for the best compatible version. We're not there yet and this might be irrelevant to this discussion right now, but I thought I'd share this vision as I've been thinking about Substrate versioning a lot recently 😅

I'm asking for semver for 2+ years :D so hopefully

PR #610, opened 2023-04-19
chore: substrate upgrade to v0.9.38

Roznovjak:
Based on #607.

Notable changes:

- renaming in pallets
  - Event -> RuntimeEvent
  - Origin -> RuntimeOrigin
  - Call -> RuntimeCall
- extrinsics need call_index
- orml_tokens - callbacks OnDust,OnNewAccoutn and OnKilledAccount combined into one CurrencyHooks
- storage migrations:
  - Preimage v0 -> v1
  - Democracy v0 -> v1
    - Preimages are NOT migrated. Any registered preimages in Democracy at the time of migration are dropped. Their balance is NOT UNRESERVED.
    - The re-dispatcher used in the old Democracy implementation is removed. Any proposals scheduled for dispatch by Democracy WILL NOT EXECUTE.
  
    This means you SHOULD ensure that:
    - the preimage for the runtime upgrade is placed as an imminent preimage, not with a deposit;
    - no other preimages are in place at the time of upgrade;
    - there are no other proposals scheduled for dispatch by Democracy at the time of upgrade.
  - Multisig v0 -> v1
    - Any outstanding multisig calls (ones where one signatory has signed, but not the threshold) will not be able to execute.
  - Pallet XCM v0 -> v1
  - DMP Queue v0 -> v2
  - XCMP Queue v1 -> v3
  - Parachain System v1 -> v2
  - Scheduler v0 -> v4
    - The scheduler will drop all un-decodable Agendas. This can happen in case a Call in an Agenda becomes un-decodable.
The migration will print how many Agendas have been dropped.
  - Duster v0 -> v1
  - AssetRegistry - pallet version is not increased

TODO: 
- rebenchmark


apopiak:
> Scheduler v0 -> v4

That seems dicey? What's our strategy there?

Roznovjak:
> > Scheduler v0 -> v4
> 
> That seems dicey? What's our strategy there?

Do you see any issue with it? There is a specific method for it (`migrate_v1_to_v4`), so I wouldn't expect any issue. Moreover, it's the only available method. Other methods require higher storage versions.

apopiak:
Never mind then, all good 👍

runtime/common/src/lib.rs
```diff
@@ -132,7 +133,7 @@ parameter_types! {
 	/// Minimum amount of the multiplier. This value cannot be too low. A test case should ensure
 	/// that combined with `AdjustmentVariable`, we can recover from the minimum.
 	pub MinimumMultiplier: Multiplier = Multiplier::saturating_from_rational(1, 1_000_000u128);
-	pub const MultiPaymentCurrencySetFee: Pays = Pays::Yes;
+	pub MaximumMultiplier: Multiplier = Bounded::max_value();
```
mrq1911, line 136: shouldn't be this set to less than impossible to pay value? so it is still somewhat possible to interact with chain and not completely out of reach to anyone? especially when "free" extrinsic that still take weight can drive the price up

primitives/src/constants.rs
```diff
@@ -89,7 +89,10 @@ pub mod chain {
 	pub const MIN_POOL_LIQUIDITY: Balance = 1000;
 
 	/// We allow for
-	pub const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_ref_time(WEIGHT_PER_SECOND.ref_time() / 2);
+	pub const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
+		WEIGHT_REF_TIME_PER_SECOND.saturating_div(2),
+		polkadot_primitives::v2::MAX_POV_SIZE as u64,
```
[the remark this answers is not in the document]
Roznovjak: polkadot and kusama use u64::MAX. Acala uses what we have there right now.

runtime/common/src/lib.rs
Roznovjak: I compared the implementations and this should behave the same way and provide the same results as before. Before, the returned value was calculated as `something.max(min_multiplier)`. Now the value is clamped `something.clamp(min_multiplier, max_multiplier)`. So the new param `MaximumMultiplier` is used to restrict the multiplier, which was "unlimited" before.

mrq1911: i understand that it was unlimited before - it was definitely an issue. that is why I want to utilize this param actually and set it sensibly

mrq1911: lets fix it after merge in https://github.com/galacticcouncil/Basilisk-node/issues/613

PR #614, opened 2023-05-25
fix: XYK Spot Price

apopiak:
+ Inverts the `XYKSpotPrice` implementation in line with https://github.com/galacticcouncil/warehouse/pull/204
+ Adjusts fee integration tests.

## Checklist:

### DIFF
--- a/integration-tests/src/non_native_fee.rs
+++ b/integration-tests/src/non_native_fee.rs
@@ -39,11 +39,8 @@ fn non_native_fee_payment_works_with_xyk_spot_price() {
 	TestNet::reset();
 
 	Basilisk::execute_with(|| {
-		let currency_0 = 0;
-		let currency_1 = 1;
-
 		let call = basilisk_runtime::RuntimeCall::MultiTransactionPayment(
-			pallet_transaction_multi_payment::Call::set_currency { currency: currency_1 },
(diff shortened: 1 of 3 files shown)


integration-tests/src/non_native_fee.rs
```diff
@@ -127,8 +172,8 @@ fn non_native_fee_payment_works_with_xyk_spot_price() {
 			)
 		);
 
-		let dave_balance = basilisk_runtime::Tokens::free_balance(1, &AccountId::from(DAVE));
-		assert_eq!(dave_balance, 970_376_058_464_304);
+		let dave_balance = basilisk_runtime::Tokens::free_balance(NEW_TOKEN, &AccountId::from(DAVE));
+		assert_eq!(dave_balance, 90_264_297_166_679);
```
enthusiastmartin, line 176: this difference is quite big. how so ?

apopiak: most of the change was because I set Dave's balance to 100 instead of 1000 UNITS.
changed that now.
The fee reducing from ~30 to ~10 is not *that* unexpected because the price changed from 0.5 to 2.0 (factor of 4). I am unsure why it didn't just divide by 4, though. (30/4 = 7.5)

PR #616, opened 2023-05-30
feat: asset registry xcm rate limit

apopiak:
Add a migration for the asset registry that removes `locked` and introduces `xcm_rate_limit`.

https://github.com/galacticcouncil/warehouse/pull/209

## Checklist:


mrq1911 (approved):

apopiak:
note: migration not tested, yet

PR #633, opened 2023-07-03
feat: xcm with computed origin config

Roznovjak:
Based on a discussion with @apopiak, where we reviewed the XCM changes proposed by the Substrate builders programme in this PR #591, we decided to make the following changes:
1. Use `WithComputedOrigin` derivative barrier with `AllowTopLevelPaidExecutionFrom` and `AllowSubscriptionsFrom` barriers. This outer barrier computes the effective origin (checks for any origin altering operations) and passes this updated origin to inner barriers.
2. Change `XcmExecuteFilter` from `Everything` to `Nothing`. This change disallows `pallet_xcm::execute` extrinsic which allows sending arbitrary XCM messages. We are not aware of any issues with this param set to `Everything`, but better be sure. Moreover, `pallet_xcm` is blocked by the runtime call filter.


runtime/basilisk/src/xcm.rs
```diff
@@ -57,15 +57,17 @@ pub type LocalOriginToLocation = SignedToAccountId32<RuntimeOrigin, AccountId, R
...
-	// The barrier will only pass when the origin is a Tinkernet multisig.
 	WithComputedOrigin<
-		AllowTopLevelPaidExecutionFrom<invarch_xcm_builder::TinkernetMultisigMultiLocation>,
+		(
+			AllowTopLevelPaidExecutionFrom<Everything>,
+			// Subscriptions for version tracking are OK.
+			AllowSubscriptionsFrom<Everything>,
+			// Tinkernet Multisigs require DescendOrigin, this barrier makes sure it's followed by payment.
+			// The barrier will only pass when the origin is a Tinkernet multisig.
+			AllowTopLevelPaidExecutionFrom<invarch_xcm_builder::TinkernetMultisigMultiLocation>,
```
apopiak, line 69: These two seem redundant?

runtime/basilisk/src/xcm.rs
```diff
@@ -57,15 +57,17 @@ pub type LocalOriginToLocation = SignedToAccountId32<RuntimeOrigin, AccountId, R
...
+		),
 		UniversalLocation,
 		ConstU32<1>,
```
apopiak, line 72: Same comment as for Hydra, ecosystem seems to go for 8, no reason to deviate

apopiak (approved):
looks fine after addressing comments

runtime/basilisk/src/xcm.rs
mrq1911: @Roznovjak ?

runtime/basilisk/src/xcm.rs
Roznovjak: Not sure what exactly you mean. Removing `AllowTopLevelPaidExecutionFrom<invarch_xcm_builder...>` and moving `AllowSubscriptionsFrom` outside of `WithComputedOrigin`?

runtime/basilisk/src/xcm.rs
Roznovjak: done

runtime/basilisk/src/xcm.rs
apopiak: so just the first of your comments

Roznovjak: removed `AllowTopLevelPaidExecutionFrom<invarch_xcm_builder::TinkernetMultisigMultiLocation>` from the barier.

PR #636, opened 2023-07-12
feat!: XCM Rate Limiter

apopiak:
Add the xcm-rate-limiter pallet from https://github.com/galacticcouncil/HydraDX-node/pull/561 to Basilisk.

## Checklist:

## Todos:
- [x] run benchmarks and update weights


integration-tests/parachain-runtime-mock/src/lib.rs
```diff
@@ -228,6 +229,10 @@ impl cumulus_pallet_xcmp_queue::Config for ParachainRuntime {
 	type ControllerOriginConverter = XcmOriginToCallOrigin;
 	type PriceForSiblingDelivery = ();
 	type WeightInfo = ();
+	type ExecuteDeferredOrigin = EnsureRoot<AccountId>;
+	type MaxDeferredMessages = ConstU32<100>;
```
mrq1911, line 233: what happens to the messages if this buffer is filled?

runtime/basilisk/src/xcm.rs
```diff
@@ -253,6 +259,21 @@ impl pallet_xcm::Config for Runtime {
 	type ReachableDest = ReachableDest;
 }
 
+parameter_types! {
+	pub DeferDuration: RelayChainBlockNumber = 600; // 1 hour
```
mrq1911, line 263: seems bit long

mrq1911 (approved):
fix tests and LGTM

integration-tests/parachain-runtime-mock/src/lib.rs
apopiak: We drop the message.
https://github.com/galacticcouncil/cumulus/pull/3/files#diff-3d3febd8cfcd42ba5bb949d63fdafb9a71da3db1feb35a45601404b64fcd70c4R841

runtime/basilisk/src/xcm.rs
apopiak: change to 10min

apopiak:
Reason for this is that tests are run with `--features runtime-benchmarks` which changes the `RelayChainBlockNumberProvider` -.-

PR #644, opened 2023-10-20
feat: Update XCMP Queue to Extended Version

apopiak:
This PR updates the XCMP queue to the new version from https://github.com/galacticcouncil/cumulus/pull/4 and adjusts configuration appropriately.

## Motivation and Context

We want the queue to be able to hold more than 100 messages (the previous limit). The new limit is between 1000 and 100_000 messages (depending on the fullness of the buckets).

## Checklist:

- [ ] I have updated the documentation if necessary.
- [x] Update weights.
- [ ] I have added tests to cover my changes, regression test if fixing an issue.
- [x] This is a breaking change.

PR #648, opened 2023-12-05
new xcm rate limiter

mrq1911:


jak-pan:
Does this deprecate #644 or is it complementary?

apopiak:
> Does this deprecate #644 or is it complementary?

It deprecates it, but these changes are already in polkadot v1.1.

PR #645, opened 2023-10-20
chore: upgrade to polkadot-v1.1.0

Roznovjak:
TODO:
- [x] use XYK and LBP pallets from hydra
- [x] rebenchmark all pallets
- [x] storage migrations
- [x] test runtime upgrade (try-runtime and chopsticks)
- [x] update main README file

Our stuff:
- [x] XCMP defer queue
- [x] XCM executor atomicity
- [x] Tinkernet
- [x] Democracy pallet
- [x] rust 1.71 compatibily with Zombienet (`UnknownOpcode(192)`)

Notable changes:
- Deprecate `Weight::from_ref_time` and `Weight::from_proof_size` and use `from_parts` instead
- The `try-runtime` subcommand has been migrated to a standalone CLI [#14731](https://github.com/paritytech/substrate/pull/14731)
- [XCM Topics](https://github.com/paritytech/polkadot/pull/7234) - XCM messages are uniquely referenceable
- `BlockNumber` removed from `frame_system::Config`. Use `frame_system::BlockNumberFor<T>`. #[14437](https://github.com/paritytech/substrate/pull/14437)
- Default config parameters #[13454](https://github.com/paritytech/substrate/pull/13454)

API changes:
- pallet_balances #[12951](https://github.com/paritytech/substrate/pull/12951)
  - `set_balance` becomes `force_set_balance`
  - `transfer` becomes `transfer_allow_death`

This release removes the following CLI parameters:

    --rpc-max-payload (replaced by --rpc-max-request-size and --rpc-max-response-size)
    --ws-max-out-buffer-capacity
    --ws-external (replaced by --rpc-external)
    --unsafe-ws-external (replaced by --unsafe-rpc-external)
    --ipc--path
    --ws-port (replaced by --rpc-port)
    --ws-max--connections (replaced by --rpc-max-connections)
    --rpc-http (replaced by --rpc-addr)
    --rpc-ws (replaced by --rpc-addr)

PR #652, opened 2024-01-02
ci: add write permission for version-check job

Roznovjak:
Should fix the error from version-check job:
Error: Resource not accessible by integration


mrq1911:
replaced by #645

PR #654, opened 2024-01-29
feat: state trie migration

Roznovjak:
This PR integrates the state trie migration pallet and sets the state version of the runtime to 1. This change enables lazy storage migration.


runtime/basilisk/src/system.rs
```diff
@@ -468,3 +469,35 @@ impl pallet_multisig::Config for Runtime {
 	type MaxSignatories = MaxSignatories;
 	type WeightInfo = ();
 }
+
+pub struct TechCommAccounts;
+impl SortedMembers<AccountId> for TechCommAccounts {
```
mrq1911, line 474: please use the tech committee storage for this

node/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "basilisk"
-version = "11.0.0"
+version = "11.0.1"
```
mrq1911, line 3: this is at least a minor bump

mrq1911 (changes requested):

jak-pan:
Wait do we also need this on snek?

Roznovjak:
> Wait do we also need this on snek?

AFAIK we have the issue with PoV size on Basilisk as well.

mrq1911 (approved):

PR #655, opened 2024-02-15
Add HashedDescription XCM converter and remove TinkernetMultisig configs

arrudagates:
This PR adds the common XCM location converter HashedDescription, which is used across the ecosystem, to Basilisk.

Since the current Polkadot SDK version in use does not contain the `DescribeBodyTerminal` config, it is also temporarily pulled locally into `xcm.rs`, to be removed once the runtime is upgraded to polkadot-v1.2.0, which [adds it to DescribeAllTerminal](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.2.0/polkadot/xcm/xcm-builder/src/location_conversion.rs#L104).

For reference, chains using HashedDescription include the [relay chain](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/xcm_config.rs#L76) and system parachains like the [Asset Hub](https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs#L93).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -900,7 +900,6 @@ dependencies = [
  "orml-unknown-tokens",
  "orml-vesting",
  "orml-xcm",
- "orml-xcm-builder-kusama",
  "orml-xcm-support",
  "orml-xtokens",
  "pallet-asset-registry",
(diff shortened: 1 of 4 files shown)


arrudagates:
@green-jay Any other blockers on this?

PR #662, opened 2024-05-14
fix: runtime pallet order

enthusiastmartin:


mrq1911 (approved):

Roznovjak (commented):
Please also update the order of on_initialize/finalize in integration_tests/polkadot_test_net.rs

enthusiastmartin:
> Please also update the order of on_initialize/finalize in integration_tests/polkadot_test_net.rs

this does not apply to basilisk. there is not a particular place where we can on_initialize.

PR #663, opened 2024-05-29
feat: OpenGov

vgantchev:
This PR adds OpenGov which for some time will run in parallel to gov v1.

Tracks:
https://docs.google.com/spreadsheets/d/1udPnZlojVjawF8oXc2fKhZTZNZdthAusgOIFtJs9tA8/edit#gid=0

Origins:
https://docs.google.com/spreadsheets/d/1uZKwzy4vv1W10XslKcyRVOBPX0Pm7PcWKpSoG36Xvh0/edit#gid=0


primitives/src/constants.rs
```diff
@@ -21,6 +21,7 @@ pub mod currency {
 	pub const FORTUNE: Balance = u128::MAX;
 	pub const UNITS: Balance = 1_000_000_000_000;
 	pub const DOLLARS: Balance = UNITS * 100; // 100 UNITS ~= 1 $
+	pub const GRAND: Balance = UNITS * 1_000;
```
dmoka, line 24: isn't a grand equal to a thousand of dollars? So it should be `DOLLARS * 1_000`?!

mrq1911: yeah, especially looking at usage of it i think it should be DOLLARS * 1000

vgantchev: I never understood the concept of DOLLARS
Should it be some approximation of $1 worth of BSX? Then it's wrong anyway (BSX price is 0.00009)

mrq1911: originally it was approximation of the approximation of the price of BSX compared to HDX. if the limits should be in BSX, don't use confusing dollars nomenclature and just use UNITS

mrq1911: but GRAND should be 1000 DOLLARS same as CENT being DOLARS / 100

mrq1911 (approved):

PR #667, opened 2024-08-30
fix: chain spec and refactor zombienet

vgantchev:
This PR fixes the chain spec by migrating it from `from_genesis` to `GenesisBulder`

It also updates the Zombienet config for the native local testnet and removes unused configs


README.md
```diff
@@ -30,6 +30,17 @@ make build
...
+* Relay chain repository (polkadot) has to be built in `../polkadot`
+* Grab `zombienet` utility used to start network from [releases](https://github.com/paritytech/zombienet/releases)
+
+
+```
+cd ./zombienet
+zombienet spawn local.json
+```
+
 ### Local Testnet
```
mrq1911, line 44: delete as you deleted config

zombienet/local.json
```diff
@@ -0,0 +1,65 @@
+{
```
mrq1911, line 1: i would move this to launch-configs folder

zombienet/local.json
```diff
@@ -0,0 +1,65 @@
...
+  "settings": {
+    "timeout": 1000,
+    "provider": "native"
+  },
+  "relaychain": {
+    "default_command": "../../polkadot/target/release/polkadot",
+    "chain": "rococo-local",
+    "default_args": [
+      "--no-hardware-benchmarks",
+      "-lparachain=debug",
```
mrq1911, line 11: no debug pls by default

node/src/chain_spec/rococo.rs
```diff
@@ -0,0 +1,98 @@
+// This file is part of Basilisk-node
```
mrq1911, line 1: this file shouldnt be needed anymore, rococo chainspec was already generated

node/src/chain_spec/paseo.rs
```diff
@@ -0,0 +1,98 @@
...
+	);
+
+	let chain_spec = ChainSpec::builder(
+		wasm_binary,
+		Extensions {
+			relay_chain: "rococo-local".into(),
+			para_id: PARA_ID,
+		},
+	)
+	.with_name("Basilisk Testnet (Rococo)")
```
mrq1911, line 88: Paseo, but really this should be used only once to generate raw chainspec for paseo, then when you specify chain=paseo, it should use generated raw chainspec (same as prod) instead of generating fresh one

mrq1911: this could be paseo-local if you want to keep both chains

node/src/chain_spec/rococo.rs
mrq1911: chain=rococo should use that generated raw chainspec same as prod

mrq1911 (changes requested):

mrq1911 (approved):

PR #670, opened 2024-09-17
feat: Sunset gov v1

vgantchev:
This PR sunsets gov v1 and the Council, more specifically:
* Removes the following pallets from the runtime: `elections-phragmen`, `tips`, `collectives/council`
* Adds migrations to the locks on funds and clean the storage associated with the pallets above

`pallet-democracy` is left in the runtime for now, to be cleaned up later once the existing votes have expired (another max ~3 months)


runtime/basilisk/src/lib.rs
```diff
@@ -170,13 +171,9 @@ construct_runtime!(
 		Treasury: pallet_treasury = 4,
 		Utility: pallet_utility = 5,
 		//NOTE: 6 - is used by Scheduler which must be after cumulus_pallet_parachain_system
-		Democracy: pallet_democracy exclude_parts { Config } = 7,
```
mrq1911, line 173: we should probably retire those pallet id's of removed pallets

mrq1911 (approved):
fmt & bump versions & LGMT

runtime/basilisk/src/lib.rs
vgantchev: wdym by retire the pallet IDs?

mrq1911: not use them in future

mrq1911 (approved):

runtime/basilisk/src/lib.rs
```diff
@@ -171,12 +172,9 @@ construct_runtime!(
 		Utility: pallet_utility = 5,
 		//NOTE: 6 - is used by Scheduler which must be after cumulus_pallet_parachain_system
 		Democracy: pallet_democracy exclude_parts { Config } = 7,
-		Elections: pallet_elections_phragmen = 8,
```
jak-pan, line 174: We should check the storage after we remove pallets, I think we'll need to clear it either by ref or by migration.

runtime/basilisk/src/governance/old.rs
```diff
@@ -19,29 +19,12 @@ use crate::governance::TechnicalCollective;
 use crate::*;
 
 use primitives::constants::{
-	currency::{CENTS, DOLLARS, UNITS},
+	currency::DOLLARS,
```
jak-pan, line 22: Why remove this and then use imports directly? Just change to primitives if it changed?

runtime/basilisk/src/lib.rs
jak-pan: Ok, I see later we do migration

PR #672, opened 2024-10-02
feat: 6 second block time

vgantchev:
This PR introduces async backing (6s block time)

The changes have been prepared following this guide: https://wiki.polkadot.network/docs/maintain-guides-async-backing


runtime/basilisk/src/lib.rs
```diff
@@ -359,30 +360,7 @@ mod benches {
 	);
 }
```
jak-pan, line 362: I think we need the inherents. ( I saw somebody had problem with this )

jak-pan: But if we have it elsewhere then ok

node/src/service.rs
```diff
@@ -417,12 +424,11 @@ fn start_consensus(
 		relay_chain_slot_duration,
 		proposer,
 		collator_service,
-		// Very limited proposal time.
-		authoring_duration: Duration::from_millis(500),
-		collation_request_receiver: None,
+		authoring_duration: Duration::from_millis(1500),
```
jak-pan, line 427: This doesn't mean execution time right? It's just how much time we have to propose a block

runtime/basilisk/src/system.rs
```diff
@@ -414,11 +423,19 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 	type ReservedDmpWeight = ReservedDmpWeight;
 	type XcmpMessageHandler = XcmpQueue;
 	type ReservedXcmpWeight = ReservedXcmpWeight;
-	type CheckAssociatedRelayNumber = cumulus_pallet_parachain_system::RelayNumberStrictlyIncreases;
+	type CheckAssociatedRelayNumber = cumulus_pallet_parachain_system::RelayNumberMonotonicallyIncreases;
 	type DmpQueue = frame_support::traits::EnqueueWithOrigin<MessageQueue, RelayOrigin>;
+	type ConsensusHook = ConsensusHook;
```
jak-pan, line 428: I believe this is where inherents are checked now?

jak-pan (approved):

runtime/basilisk/src/lib.rs
jak-pan: I'll answer myself, it's in auraExt now

runtime/basilisk/src/system.rs
jak-pan: Yes...

primitives/src/constants.rs
```diff
@@ -88,9 +75,9 @@ pub mod chain {
 	/// Minimum pool liquidity
 	pub const MIN_POOL_LIQUIDITY: Balance = 1000;
 
-	/// We allow for
+	/// We allow for 2 seconds of compute with a 6 second average block.
 	pub const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
-		WEIGHT_REF_TIME_PER_SECOND.saturating_div(2),
+		WEIGHT_REF_TIME_PER_SECOND.saturating_mul(2),
```
mrq1911, line 80: size of block (weight) is 4x what was before @jak-pan

runtime/basilisk/src/system.rs
```diff
@@ -507,7 +524,7 @@ impl pallet_transaction_pause::Config for Runtime {
 }
 
 parameter_types! {
-	pub const RewardPerCollator: Balance = 15_216_000_000_000_000; // 12.68[BSX/block] * 1200[block]
+	pub const RewardPerCollator: Balance = 15_240_000_000_000_000; // 6.35[BSX/block] * 2400[block]
```
mrq1911, line 527: should be 1/2 if the session length remain the same (1200 blocks)

primitives/src/constants.rs
```diff
@@ -39,32 +39,19 @@ pub mod currency {
...
-	/// <https://research.web3.foundation/en/latest/polkadot/BABE/Babe/#6-practical-results>
-
-	pub const MILLISECS_PER_BLOCK: u64 = 12000;
+	/// Change this to adjust the block time.
+	pub const MILLISECS_PER_BLOCK: u64 = 6000;
+	pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
 
 	// Time is measured by number of blocks.
 	pub const MINUTES: BlockNumber = 60_000 / (MILLISECS_PER_BLOCK as BlockNumber);
 	pub const HOURS: BlockNumber = MINUTES * 60;
```
mrq1911, line 52: session length uses this constant, and that shouldn't probably change as it would break conversion from block number to session

mrq1911 (changes requested):
ensure that session pallet doesnt break

mrq1911 (approved):

PR #683, opened 2025-09-17
feat: 2s blocktime

enthusiastmartin:
This PR enables 2s blocktime (elastic scaling)


lexnv (approved):

runtime/basilisk/src/system.rs
```diff
@@ -242,7 +242,7 @@ impl pallet_timestamp::Config for Runtime {
 	/// A timestamp: milliseconds since the unix epoch.
 	type Moment = u64;
 	type OnTimestampSet = ();
-	type MinimumPeriod = ConstU64<{ SLOT_DURATION / 2 }>;
+	type MinimumPeriod = ConstU64<0>;
```
mrq1911, line 245: is this safe? i guess you are still limited by 3 blocks per slot?

runtime/basilisk/src/system.rs
```diff
@@ -611,7 +611,7 @@ impl pallet_transaction_pause::Config for Runtime {
 }
 
 parameter_types! {
-	pub const RewardPerCollator: Balance = 15_240_000_000_000_000; // 6.35[BSX/block] * 2400[block]
+	pub const RewardPerCollator: Balance = 15_240_000_000_000_000; // fixed reward per session
```
mrq1911, line 614: sessions will be 3x as often so this is effectively 3x collator rewards

cl0w5: The session length is the same - session `Period`is defined using `HOURS` const which builds upon `MILLISECS_PER_BLOCK`

see https://github.com/galacticcouncil/Basilisk-node/blob/master/runtime/basilisk/src/system.rs#L590

runtime/basilisk/src/system.rs
cl0w5: parity uses the same, see https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs

The amount of blocks per slot is limited by `BLOCK_PROCESSING_VELOCITY = 3`

